PR #10, opened 2015-09-28
Remove FOSUserBundle dependency

djandyr:
Removed FOSUserBundle dependency

- All references to FOSUserBundle extracted.
- To authenticate a user when refreshing a token, PreAuthenticatedToken is used with associated Authenticator and Provider. The pre-authentication mechanism works well, as a valid refresh token is always authenticated, a user does not need to supply a username or password. 
- Extracted RefreshTokenController into service to allow for custom routing and unit testing.
- Add config parameter to specify firewall, this is needed to identify the correct user provider.
- Removed RefreshTokenListener - maybe custom to our requirements due to a solely REST API.
- When a token is refreshed, the TTL is only increased - the token stays the same. Previously the token was changed each time.


markitosgv:
Hi @djandyr,

First of all, nice work!! I do some changes in https://github.com/gesdinet/JWTRefreshTokenBundle/tree/warrantgroup-extract-fosuserbundle-dependency

Can you check it out and then fix you PR?

I've made changes:

- Fix styling
- README
- Changes in configuration

Thanks

chalasr:
Hi. I'm implementing your bundle using FOSUserBundle as user provider.
I have created a custom JWTResponseListener to return user informations (like id) with tokens.

My problem is : the user provider called on AuthenticationSuccess don't know the `getId()`method. (Symfony\Component\Security\Core\User\User used)

How can I use FOSUserBundle interface ?
Thank's

PR #25, opened 2016-01-07
Add configuration option to expand refresh token ttl

jdecool:
To improve application security, I want to set a small refresh token TTL.

This PR allow to update the refresh token TTL after each refresh request. This that option, we can define a smaller TTL which allow user to renew his JWT token as long his session is valid.


markitosgv:
@jdecool great job, thanks

PR #26, opened 2016-01-07
Force new object to set expiration date to avoid fix error on some Doctrine version

jdecool:
We notice that on some Doctrine version, the expiration date is not modify in database.

This is the fix, to force a new DateTime object that will be considered by Doctrine for the update.

This is related to previous PR #25 .
Really sorry for the inconvenience.


jdecool:
Thanks

markitosgv:
Thanks to you for your collaboration @jdecool

PR #30, opened 2016-01-21
service parameters are now quoted according to YAML specifications

danielkay:
More recent versions of PHPUnit trip up over un-quoted '@' characters in YAML files - simply quoting service arguments fixes this and doesn't appear to cause any untoward behaviour as a result.


danielkay:
I suspect this is failing due to commits to master after the v0.1.3 tag...

markitosgv:
Can you fix this PR?

danielkay:
This works on top of the v0.1.3 release, but someone else's commits to master after that release seem to be breaking three build from what I can tell...

danielkay:
breaking the** build

markitosgv:
Ok let me revert last commit

danielkay:
thanks :)

PR #31, opened 2016-01-28
Raise minimum Symfony version to 2.5

ChristopherMoll:
ValidatorInterface used in AttachRefreshTokenOnSuccessListener was moved and deprecated in 2.5: https://github.com/symfony/validator/blob/master/CHANGELOG.md

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
   ],
   "require" : {
     "php": ">=5.3.3",
-    "symfony/symfony": ">=2.5",
+    "symfony/symfony": ">=2.5,<2.8",
     "doctrine/orm": "^2.4.8",
     "doctrine/doctrine-bundle": "~1.4",
     "lexik/jwt-authentication-bundle": "^1.1"
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
   ],
   "require" : {
     "php": ">=5.3.3",
-    "symfony/symfony": "~2.3",
+    "symfony/symfony": ">=2.5",
     "doctrine/orm": "^2.4.8",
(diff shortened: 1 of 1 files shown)


markitosgv:
I think we have to add support to Symfony 3.0. In actual branch i want to support from 2.3.

ChristopherMoll:
Sorry, I think you might have misunderstood my comment.  It doesn't currently work with Symfony 2.3 or 2.4, because in 2.5 they moved `Validator/Validator/ValidatorInterface` to `Validator/ValidatorInterface`.  This library uses that new interface, so an error is thrown when `AttachRefreshTokenOnSuccessListener` receives the deprecated interface in `__construct()`.  

The composer.json requires Symfony 2.3+, but this is inaccurate given the incompatibility above.  Since you say the other branch is going to be 2.5-3.0 anyway, this fix is really only relevant in the meantime.  But currently any 2.3 project that updates from [0.1.1](https://github.com/gesdinet/JWTRefreshTokenBundle/blob/v0.1.1/EventListener/AttachRefreshTokenOnSuccessListener.php) to [0.1.2](https://github.com/gesdinet/JWTRefreshTokenBundle/blob/v0.1.2/EventListener/AttachRefreshTokenOnSuccessListener.php) of this library will break despite composer allowing it.

markitosgv:
You're right but you need to edit composer.json to avoid 3.x versions

ChristopherMoll:
Ah, I was the one that misunderstood.  It looks like travis is failing due to [an API change made in 2.8](http://symfony.com/doc/current/cookbook/security/api_key_authentication.html#the-api-key-authenticator).

PR #34, opened 2016-02-03
Workaround to support implementing SimplePreAuthenticatorInterface from Symfony <2.7 and >2.8

dragosprotung:


dragosprotung:
Of course StyleCI will fail, but the file could be ignored until support for Symfony 2.7 is dropped.

PR #39, opened 2016-03-29
allow to extends Gesdinet\JWTRefreshTokenBundle\Entity\RefreshToken …

pitpit:
... keeping ORM base annotations working.

When overriding Gesdinet\JWTRefreshTokenBundle\Entity\RefreshToken with another class (to customize the table name actually), extending Gesdinet\JWTRefreshTokenBundle\Entity\RefreshToken because all class properties are private. And I don't want to copy&paste the whole code


laupiFrpar:
:+1:

AlexStef:
:+1:

PR #45, opened 2016-07-23
Use RequestStack rather than deprecated Lexik *Event::getRequest()

chalasr:
As we are deprecating all Request instances injection from our Event classes (see https://github.com/lexik/LexikJWTAuthenticationBundle/pull/200) in order to totally remove them in `lexik/jwt-authentication-bundle:2.0` (see https://github.com/lexik/LexikJWTAuthenticationBundle/pull/201), this makes this bundle working accordingly.

It would be nice to quickly move on this, as the `lexik/jwt-authentication-bundle` v2.0 will soon be released and this bundle (that is quite important for us) would not work as is.


markitosgv:
Of course, test fails because memory issue. Merged it. Thank you!

PR #48, opened 2016-10-01
Add a config node to change the property name of refresh token

philippecarle:
Instead of having the "refresh_token" property hard-coded into the listener, I think it would be a good practice to use the configuration with a default value. Developers could use it to decide how to name the json node.


philippecarle:
Is this repository not maintained anymore ?

PR #55, opened 2017-02-06
use minimal symfony dependencies

fbourigault:
This allows to not install the whole symfony/symfony dependency as only symfony/framework-bundle and symfony/validator are required.

Required packages were found using `grep -Rh --exclude-dir=vendor --exclude-dir=bin --include='*.php' 'Symfony' | sort | uniq`.

Vendor directory is now 40MB instead of 63MB!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
   ],
   "require" : {
     "php": ">=5.3.3",
-    "symfony/symfony": "~2.3|~3.0",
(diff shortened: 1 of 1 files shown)


markitosgv:
Thanks and merged!

fbourigault:
Is it possible to have a release which includes this PR? Maybe a `0.2.0` which looks to me the safest to not break BC.

markitosgv:
Done! And thanks again!

PR #56, opened 2017-03-15
Use a UserChecker

### LINKED ISSUE #52: Security issue: Loadable users are always authenticated
The current implementation to refresh JWT tokens is not safe. When a user presents a refresh token, the only check that occurs is whether the user provider can load the user. If a user is loadable, it is granted a new access token.

But the authenticator should also check if the loaded user is still allowed to access the application. For example, users that implement the AdvancedUserInterface should be checked whether their account is enabled, is not locked and is not expired. Other means of authentication may have other requirements (e.g. for OAuth accounts, check whether the correct permissions are still present).

Currently, this bundle will happily give out new access tokens to users who have been disabled, expired, etc.

I think there should be some way to hook into the authenticator to provide this logic. Or perhaps the authenticator in this bundle should delegate authentication to some other authentication provider.

Currently I am using a compiler pass to overwrite the authenticator provided in this bundle with my own. That's a bit of a hackish solution.

sandermarechal:
Use a UserChecker during authentication. This fixes #52

### DIFF
--- a/DependencyInjection/Compiler/UserCheckerCompilerPass.php
+++ b/DependencyInjection/Compiler/UserCheckerCompilerPass.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Gesdinet\JWTRefreshTokenBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * UserCheckerCompilerPass.
+ */
+final class UserCheckerCompilerPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $userCheckerId = $container->getParameter('gesdinet.jwtrefreshtoken.user_checker.id');
+        if (!$userCheckerId) {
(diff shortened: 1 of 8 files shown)


sandermarechal:
Perhaps the documentation should even go as far as recommending (or requiring?) to set a user provider? Not doing so is a security issue. The default user checker checks for inactive or locked accounts, but users generated by this bundles user_provider generally pass all those checks, even when the should not.

juliusstoerrle:
What's the plan for this merge request? We need this feature in order to use this Bundle and would appreciate if those changes would be merged.

lashae:
Is there any progress or ETA about merging this PR?

xthiago:
I'm guess the project was abandoned. The author does not reply any pull-request. I'm considering creating a fork project.

markitosgv:
@xthiago Im searching for some collaborators to help me with this repository, do you want to participate?

markitosgv:
@sandermarechal colud you please up to date this PR and resolve conflicts for merging? thanks

sandermarechal:
I have rebased on master and resolved all the conflicts. Only, I could not run the tests locally because I do not have the mongodb extension installed and composer refuses to install. Perhaps the mongodb extension should be optional for testing, just skipping the tests?

PR #63, opened 2017-04-10
Set authention token after authenticating with refresh token

sandermarechal:
This PR sets the correct security token in the token_storage after authenticating with a refresh token. This allows services that run after authentication (e.g event handlers that listen to onAuthenticationSuccess) to use Symfony services like the authorization_checker.

### DIFF
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -14,7 +14,7 @@ services:
 
     gesdinet.jwtrefreshtoken:
         class: Gesdinet\JWTRefreshTokenBundle\Service\RefreshToken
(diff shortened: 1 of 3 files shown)


xthiago:
@sandermarechal the tests on Travis CI was failed. Could you take a look?

sandermarechal:
I'm not quite sure how to fix this. My patch uses the TokenStorageInterface which was introduced in Symfony 2.6. In order to make it work on Symfony 2.5 I would need to inject the entire SecurityContext. I don't know how to inject different versions based on the Symfony version or properly typehint on them.

Seeing as how support for Symfony 2.5 ended in July 2015, perhaps consider dropping support for Symfony 2.5 (and 2.6) from this bundle? At the same time we could add Symfony 3.1-3.3 and PHP 7.1 which are now missing from the Travis build matrix.

PR #69, opened 2017-07-24
Match full variety of Content-Type values

sh41:
At the moment the "Content-Type" header must be exactly 'application/json' for the refresh token to be extracted from the request when it is in json format.

It is valid for "Content-Type" header to have a parameter for example 'application/json;charset=UTF-8', but this does not match the criteria and so the refresh fails. (https://www.w3.org/Protocols/rfc1341/4_Content-Type.html)

This change borrows logic from here: https://github.com/symfony/symfony/blob/4427cf9157d5d896e2221a31afeac293f2ad9971/src/Symfony/Component/Security/Http/Firewall/UsernamePasswordJsonAuthenticationListener.php#L79 in order to check if the incoming request is in json format. 

This means that content types with parameters will be accepted. Also both 'application/json' & 'application/x-json' are considered by Symfony to be json, this is the code that determines it: https://github.com/symfony/symfony/blob/4bd7b921f4047fa32478459a9b50e63001372aee/src/Symfony/Component/HttpFoundation/Request.php#L1851

### DIFF
--- a/Request/RequestRefreshToken.php
+++ b/Request/RequestRefreshToken.php
@@ -18,7 +18,7 @@ class RequestRefreshToken
     public static function getRefreshToken(Request $request)
     {
         $refreshTokenString = null;
-        if (0 === strpos($request->getContentType(), 'application/json')) {
+        if (false !== strpos($request->getContentType(), 'json')) {
             $content = $request->getContent();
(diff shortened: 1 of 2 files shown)


markitosgv:
Nice, than you @sh41 !

PR #71, opened 2017-07-26
Create AbstractRefreshToken to allow custom RefreshToken entities

### LINKED ISSUE #51: Cannot use another table name
I have custom entity which extends from the `RefreshToken` class. When updating the schema, I get two tables: `refresh_tokens` and `myname_refresh_tokens`.

### LINKED ISSUE #70: It's not possible to override the default RefreshToken class
I want to extend the default implementation of `RefreshToken` in order to add some attributes. If I configure my class following the documentation I got the following error:

```
[Doctrine\DBAL\Exception\InvalidFieldNameException]
An exception occurred while executing 'SELECT r0_.id AS id_0, r0_.refresh_token AS refresh_token_1, r0_.username AS username_2, r0_.valid AS valid_3, c1_.custom_attr AS
impersonator_4 FROM acme_jwt_refresh_token c1_ LIMIT 1':
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'r0_.id' in 'field list'



[Doctrine\DBAL\Driver\PDOException]
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'r0_.id' in 'field list'



[PDOException]
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'r0_.id' in 'field list'
```
It happens both on refresh_token endpoint or when executing a query like `bin/console doctrine:query:dql "SELECT rf FROM Acme:RefreshToken\RefreshToken rf" --max-result=1`

In order to fix it the bundle must have an `@MappedSuperclass` class.

I will open a pull-request with the fix soon.

Aditional info:

config.yml
```yml
gesdinet_jwt_refresh_token:
    refresh_token_entity: 'Acme\Model\RefreshToken\RefreshToken'
```

RefreshToken.php
```php
<?php

declare(strict_types = 1);

namespace Acme\Model\RefreshToken;

use Gesdinet\JWTRefreshTokenBundle\Entity\RefreshToken as BaseRefreshToken;

class RefreshToken extends BaseRefreshToken
{
    /**
     * @var string
     */
    protected $customAttr;

    public function customAttr() : ?string
    {
        return $this->customAttr;
    }

    public function setcustomAttr(string $customAttr) : void
    {
        $this->customAttr = $customAttr;
    }
}

```

Doctrine mapping:
```xml
<?xml version="1.0" encoding="utf-8"?>
<doctrine-mapping
  xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping https://raw.github.com/doctrine/doctrine2/master/doctrine-mapping.xsd">

  <entity
    name="Acme\Model\RefreshToken\RefreshToken"
    table="acme_jwt_refresh_token"
    repository-class="Gesdinet\JWTRefreshTokenBundle\Entity\RefreshTokenRepository">

    <field name="customAttr" column="custom_attr" type="string" length="255" nullable="true" />

</entity>
</doctrine-mapping>
```

xthiago:
This PR fixes #70.

In the current implementation, if someone creates a custom `EntityToken` entity, the table `refresh_tokens` will also be created but no used. If you want I can provide a compiler pass to add the Doctrine mapping for `RefreshToken` based on `gesdinet_jwt_refresh_token.refresh_token_entity`. It will fix #51.

### DIFF
--- a/Entity/AbstractRefreshToken.php
+++ b/Entity/AbstractRefreshToken.php
@@ -1,0 +1,152 @@
+<?php
+
+/*
+ * This file is part of the GesdinetJWTRefreshTokenBundle package.
+ *
+ * (c) Gesdinet <http://www.gesdinet.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gesdinet\JWTRefreshTokenBundle\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Validator\Constraints as Assert;
+use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenInterface;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
+
+/**
+ * Abstract Refresh Token.
+ *
+ * @ORM\MappedSuperclass()
+ * @UniqueEntity("refreshToken")
+ */
+abstract class AbstractRefreshToken implements RefreshTokenInterface
+{
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(name="refresh_token", type="string", length=128, unique=true)
+     * @Assert\NotBlank()
+     */
+    private $refreshToken;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(name="username", type="string", length=255)
+     * @Assert\NotBlank()
+     */
+    private $username;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(name="valid", type="datetime")
+     * @Assert\NotBlank()
+     */
+    private $valid;
+
+    /**
+     * {@inheritdoc}
+     */
+    abstract public function getId();
+
+    /**
+     * Set refreshToken.
+     *
+     * @param string $refreshToken
+     *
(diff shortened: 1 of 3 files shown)


xorik:
@xthiago is it possible to remove `@UniqueEntity`, or maybe all doctrine annotations, so it will be possible to extend your abstract to a mongodb model (see #76)

validumitruro:
This might work also for the upcoming Symfony 4, which has the exact same issue (well, 3.3 does). If it's a feasible solution, when can we expect this to be merged into the master branch? At the moment, it doesn't seem possible to use custom entities at all due to the MappedSuperclass problem. (same SQL error regarding the r0_.id column) Also, will custom repositories also work with this approach?

Many thanks for the contribution, by the way.

xthiago:
> @xthiago is it possible to remove @UniqueEntity, or maybe all doctrine annotations, so it will be possible to extend your abstract to a mongodb model (see #76)

I think so.

I'm waiting any feedback from project maintainers. What you think @markitosgv?

PR #80, opened 2017-12-12
Added missing $providerKey property & type hints

toby-griffiths:
Nothing too I noticed that a property was missing, so I added it.

I also added the property type hints while I was there.


toby-griffiths:
Could you give me some clue as to why you just closed this, @markitosgv ?

markitosgv:
Please update this PR to current master branch so I could merge it

toby-griffiths:
OK.  No problem.  Will do, when I have a moment.

PR #83, opened 2018-01-14
Force

Doss:


EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -21,17 +21,30 @@
...
+    protected $userManager;
 
-    public function __construct(RefreshTokenManagerInterface $refreshTokenManager, $ttl, ValidatorInterface $validator, RequestStack $requestStack)
+    public function __construct(
+        RefreshTokenManagerInterface $refreshTokenManager,
+        int $ttl,
+        ValidatorInterface $validator,
+        RequestStack $requestStack,
+        string $getUsername,
+        UserManager $userManager
```
norkunas: This is a hard dependency on FOSUserBundle but you should consider that not all uses it

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -46,14 +59,18 @@ public function attachRefreshToken(AuthenticationSuccessEvent $event)
 
         $refreshTokenString = RequestRefreshToken::getRefreshToken($request);
 
+        $user->setLastLogin(new \DateTime());
```
norkunas: This will break if app user doesn't have this method as symfony UserInterface doesn't have this method

PR #86, opened 2018-01-17
Fix command for SF4

Heer0:
Because of missing "public: true" to "refresh_token_manager", commands does not works with sf4.

This PR fix it.


norkunas:
This missed adding `public: true` also on `gesdinet.jwtrefreshtoken`. So the service is not public and is removed from container and cannot be used for routing

Heer0:
it's been added to this PR

https://github.com/gesdinet/JWTRefreshTokenBundle/pull/87

PR #93, opened 2018-01-26
Drop dead code

norkunas:
Symfony `Request::get()` already covers getting token from the body so this part is never reached.


markitosgv:
@norkunas exaclty, get shortcut works on every parameter bag

PR #102, opened 2018-05-22
return support mongodb

anyx:
This pr not provide full support of mongodb: there is not a mongo document mapping. Maybe later i can finish it. But maybe now it will help someone to install it.


markitosgv:
Solved in #108

PR #108, opened 2018-09-11
support Doctrine's MongoDB ODM

juliangut:
Adds support for Doctrine's MongoDB ODM out of the box

Currently is quite easy to use MongoDB ODM with this package, refresh token and repository implementations and a few config changes but the big problem is located [here](https://github.com/gesdinet/JWTRefreshTokenBundle/blob/master/GesdinetJWTRefreshTokenBundle.php#L20-L22). The package already requires "doctrine/doctrine-bundle" so that condition is always true and you will have ORM configured even in the case you don't need it

After this PR the package still defaults to Doctrine's ORM entities implementation but it's absurdly simple to switch to ODM thanks to a new configuration parameter:

```yaml
gesdinet_jwt_refresh_token:
    manager_type: mongodb
```

The trade off is that composer.json does not require "doctrine/orm" and "doctrine/doctrine-bundle" anymore. You should require them explicitly in your project (or "doctrine/mongodb-odm-bundle if you want to use ODM) 

#### Deprecations

`refresh_token_entity` config has been renamed to `refresh_token_class`
`entity_manager` has been renamed to `object_manager`

Before:
```yaml
gesdinet_jwt_refresh_token:
    refresh_token_entity: MyBundle\JwtRefreshToken
    entity_manager: my.specific.entity_manager.id
```

After:
```yaml
gesdinet_jwt_refresh_token:
    refresh_token_class: MyBundle\JwtRefreshToken
    object_manager: my.specific.entity_manager.id
```

#### Considerations

`Gesdinet\JWTRefreshTokenBundle\DependencyInjection\Compiler\EntityManagerCompilerPass` has been renamed to `Gesdinet\JWTRefreshTokenBundle\DependencyInjection\Compiler\ObjectManagerCompilerPass`

`Gesdinet\JWTRefreshTokenBundle\DependencyInjection\Compiler\DoctrineMappingsCompilerPass` has been updated to provide Doctrine's mappings for ORM or ODM depending on manager_type config parameter

Even though the change on configuration parameters this PR keeps full backwards compatibility as previous config parameters are used if new are not provided

This PR fully addresses issue #26 

Supersedes #102


markitosgv:
Thanks a lot for adding odm support @juliangut

PR #110, opened 2018-09-17
Check if manager_type option is defined before trying to get it in compiler pass

cezarystepkowski:
Hello,

I've encountered an issue that `Notice: Undefined index: manager_type` is thrown on compilation after changes from https://github.com/gesdinet/JWTRefreshTokenBundle/pull/108 when using ORM and when the `manager_type` is not specified in the configuration.

It's because in compiler pass default values are not set but only the configured values are taken into consideration. So the `$config['manager_type']`  is not set if I don't specify it explicitly in my configuration file.

This PR fixes it to keep the BC so that there is no need to explicitly add manager_type to the config.


markitosgv:
Thanks @cezarystepkowski

PR #138, opened 2019-02-15
Integrate cs fixer

norkunas:
Removed also useless phpdocs


markitosgv:
@norkunas we are using https://github.styleci.io/repos/42582199 this tool for checking styling

norkunas:
I see, but this would help to fix cs before even creating PR and see style rules in the repo :thinking:

norkunas:
@markitosgv if you want to avoid this approach, can I at least suggest to move style-ci configuration to the repository and start using new php features like short array? :blush:

PR #139, opened 2019-02-22
Fix #127 : two queries for one token

### LINKED ISSUE #127: 2 queries on refresh request
Why do again

screenshot

sylvainmetayer:


markitosgv:
hi @sylvainmetayer , What do you think about removing annotation from Abstract instead? Maybe some people want to do their own implementation

sylvainmetayer:
Hi @markitosgv , you're right, I'll edit the PR.

PR #141, opened 2019-04-03
Update the document to intruct how to set the configurations & update DB schema using migration

longnd:
I found the document didn't mention about using migration to create the new table but using `schema:update --force` only.
Also when talking about setting the configurations for the bundle, it assumes the user using Symfony 3 and the configurations are all in `/app/config.yml`


markitosgv:
thanks @longnd @norkunas

Issue #99, opened 2018-04-20
Can we customize the parameter "refresh_token"


vincent-dravert:
Can we customize the parameter "refresh_token" when calling the route to refresh the JWT , I would like it to be camelCase as the rest of the API I m building.

Same question about the response, to return "token" and "refreshToken" instead of "refresh_token"

Thanks
Vincent

norkunas:
This can be closed as implemented in #142

PR #142, opened 2019-04-03
Add ability to configure refresh token parameter name (#99)

sgorshkov:
In some use cases, ex. for better consistency in APIs with camelCase object fields naming.
So library could have parameter to customise refresh token field name.

- add optional parameter token_parameter_name to bundle configuration;
- update readme;
- fix phpspec tests;

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ gesdinet_jwt_refresh_token:
 
 ### Config Refresh token parameter Name
 
-You can define refresh token parameter name. Default value is refresh_token. You can change this value adding this line to your config.yml file:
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -178,6 +178,15 @@ gesdinet_jwt_refresh_token:
     firewall: api
 ```
 
+### Config Refresh token parameter Name
+
+You can define refresh token parameter name. Default value is refresh_token. You can change this value adding this line to your config.yml file:
```
markitosgv: config.yml is only for Sf 3.x version, please update this. Something like "to your config"

PR #146, opened 2019-04-12
Fix deprecation by using Guard instead of SimplePreAuthenticatorInterface

Mapsred:
Resolve the following deprecation : 
The "Gesdinet\JWTRefreshTokenBundle\Security\Authenticator\RefreshTokenAuthenticatorBase" class implements "Symfony\Component\Security\Http\Authentication\SimplePreAuthenticatorInterface" that is deprecated since Symfony 4.2, use Guard instead.


norkunas:
@Mapsred there are still some errors with code style

markitosgv:
Nice work @Mapsred, but it breaks some method signatures then I'm going to merge it and release it in v0.7 version. thanks!

PR #148, opened 2019-04-26
Added Single Use config option

fastnloud:
When a refresh token is consumed and the config option `single_use` is set to `true` the token will no longer be valid.


jeffreymoelands:
+1 nice feature!!!

PR #156, opened 2019-07-02
Remove Symfony 4.3 deprecations

acasademont:
Symfony 4.3 has deprecated a few things related to events. This patch removes these deprecations while keeping BC with previous symfony versions

### DIFF
--- a/Event/Event.php
+++ b/Event/Event.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Gesdinet\JWTRefreshTokenBundle\Event;
+
+use Symfony\Component\EventDispatcher\Event as BaseEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Contracts\EventDispatcher\Event as ContractsBaseEvent;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+
+if (is_subclass_of(EventDispatcher::class, EventDispatcherInterface::class)) {
+    class Event extends ContractsBaseEvent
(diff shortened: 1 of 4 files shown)


Event/Event.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Gesdinet\JWTRefreshTokenBundle\Event;
+
+use Symfony\Component\EventDispatcher\Event as BaseEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Contracts\EventDispatcher\Event as ContractsBaseEvent;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+
+if (is_subclass_of(EventDispatcher::class, EventDispatcherInterface::class)) {
```
norkunas: Why not just directly doing this in `Event/RefreshEvent.php` ? There are no other events.

acasademont: I thought this was a bit more elegant solution. The other method involves duplicating the code for the whole RefreshEvent class, AFAIK you can't just put a conditional in the class declaration, it needs to be te whole class code inside the `if` and `else`

Happy to change it though.

norkunas: Not so much code :) but @markitosgv has to decide

markitosgv: Thank you both! Nice work @acasademont

acasademont:
thx @markitosgv ! a new patch release would be also really appreciated :D

markitosgv:
done @acasademont

PR #159, opened 2019-07-18
Update README.md

daifabde:
app/console was moved to bin/console since symfony 3.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,11 +129,11 @@ With the next command you will create a new table to handle your refresh tokens
 
 **Symfony 3 Version:**   
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -129,11 +129,11 @@ With the next command you will create a new table to handle your refresh tokens
 
 **Symfony 3 Version:**   
 ```bash
-php app/console doctrine:schema:update --force
+php bin/console doctrine:schema:update --force
```
norkunas: I don't see now any difference with the commands below described for Symfony 4 version.

PR #164, opened 2019-08-18
[-] TECH: Only flush RefreshToken in RefreshTokenManager

MartianGreed:


norkunas:
Since Doctrine 3.0 `flush` has zero args.

MartianGreed:
That's ok, I didn't checked for doctrine compat !
I think I'll close this then !

PR #178, opened 2019-11-25
Test with Symfony 4.4

norkunas:


norkunas:
It seems that phpspec isn't compatible yet with Symfony 5.0, so adding only 4.4.

markitosgv:
thanks @norkunas

PR #189, opened 2020-05-04
Created service from RequestRefreshToken to allow override default behavior

balazscsaba2006:
Hi, @markitosgv!

This PR creates  a service from the `RequestRefreshToken` class. It will allow changing the refresh token extraction mode (cookie headers, etc.) without being restricted to a specific `Content-Type` header, and without the need to overload the `AttachRefreshTokenOnSuccessListener`.


balazscsaba2006:
@markitosgv Any feedback on the PR? Is there anything I could do to fix the Travis builds?

PR #190, opened 2020-05-13
Add new service for generate one RefreshToken in controller

CodingByJerez:


RiffFred:
Is this pull request still relevant? I actually thought about opening something similar to achieve something close to https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/7-manual-token-creation.md. 

I think this feature would be nice to have in a context where we offer an API to other system that are not Browser applications. For example a native APP that might have its own way to store and handle refresh tokens. 

If this suggest feature is still relevant i could offer to fix the pipeline issues and to add some documentation.

PR #193, opened 2020-06-14
Specific ocramius/package-versions version

markitosgv:


guilliamxavier:
Bad idea: I understand you're trying to avoid a [Travis failure](https://travis-ci.com/github/markitosgv/JWTRefreshTokenBundle/jobs/348938695), but that's requiring *exactly* `1.5.0` and will *prevent* installing/updating this bundle in a project that requires (directly or transitively) `^1.5.1` (for [fix](https://github.com/Ocramius/PackageVersions/releases/tag/1.5.1)).

So IMO you should revert this, and <s>[maybe add a
```json
  "conflict": {
    "ocramius/package-versions": ">=1.6"
  }
```
but that will still prevent using e.g. `^1.8` (for [Composer 2 support](https://github.com/Ocramius/PackageVersions/releases/tag/1.8.0)), so the true solution is rather to]</s> update `.travis.yml`: delete `--ignore-platform-reqs` and install the needed mongodb PHP extension correctly (maybe like [this](https://github.com/doctrine/mongodb/blob/1.6.4/.travis.yml#L60-L61), i.e. `pecl install -f mongodb-stable` and preprend `composer config "platform.ext-mongo" "1.6.16" &&` to `composer require alcaeus/mongo-php-adapter`)

markitosgv:
please @guilliamxavier do a PR fixing this

guilliamxavier:
@markitosgv: done, see #195

markitosgv:
thanks @guilliamxavier , merged!

Issue #106, opened 2018-09-04
Event before RefreshToken is persisted


SebastienTouze:
Hi all, 

I think it can be usefull to have an Event dispatched before the RefreshToke is persisted to add some data inthe Entity. 
My current usecase is that I want to store the userAgent of the  user's device when s/he connnect for the first time. Currently I do that listening to lexik_jwt_authentication.on_authentication_success adding my data and persisting again the  RefreshToken. But that makes 2 DB operations where only one is necessary. 

Having an event in the `Gesdinet\JWTRefreshTokenBundle\Doctrine\RefreshTokenManager::save` or `AttachRefreshTokenOnSuccessListener::attachRefreshToken` should allow to add some data to the Entity before persist. 

Regards

SherinBloemendaal:
I will create a pull request for this. I've implemented this. 

Use case:
When you have a multiple authenticators (like one for your web-app and one for your mobile-app) and you want to store the type.

luffy1727:
why is the pull request closed

shakaran:
The current master branch has a lot changes different for upcoming 2.0. If a new pull request is sent with the changes updated, it will be merged. I am closing this issue as cleaning very old from Symfony 4 with no support. Reopen or create a new one if you still want the changes. Thanks

PR #198, opened 2020-08-06
Dispatch an event when the refresh token entity is created but not persisted yet

SherinBloemendaal:
https://github.com/markitosgv/JWTRefreshTokenBundle/issues/106


luffy1727:
why did you close this pull request?

SherinBloemendaal:
@luffy1727 because some tests fail and i don't know how to fix them. Besides the library owner does not reply or anything like that so i consider it as denied.

fd6130:
> @luffy1727 because some tests fail and i don't know how to fix them. Besides the library owner does not reply or anything like that so i consider it as denied.

Maybe you can ping someone who can make decision for this PR?

SherinBloemendaal:
> > @luffy1727 because some tests fail and i don't know how to fix them. Besides the library owner does not reply or anything like that so i consider it as denied.
> 
> Maybe you can ping someone who can make decision for this PR?

Okay, i reopend the PR and i will have a deeper look at the tests today if i have time. When the tests succeed i will ping the owner.

SherinBloemendaal:
The tests are fixed. @markitosgv what do you say?

Event/RefreshTokenEvent.php
```diff
@@ -14,7 +14,7 @@
 use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenInterface;
 use Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken;
 
-class RefreshEvent extends Event
+class RefreshTokenEvent extends Event
```
maxhelias, line 17: It's backward compatibility. You should make a new class and extend it with `RefreshEvent` and trigger a deprecation on the old one (in phpdoc or construct i don't know how it's managed for this project).

SherinBloemendaal: Okay thanks!

PR #211, opened 2020-10-28
Fix domain name for Symfony 5

vibby:


vibby:
The same changes has been requested in others PR

PR #223, opened 2021-02-01
Change services config from yaml to php

ElGovanni:
As we know, YAML configuration is going to not be default in Symfony 6+, also is not required right now but cause problem for those projects which don't use symfony/yaml so that's why I'm making this PR and ask to include for new version.


markitosgv:
Hi @ElGovanni thanks, but what about .xml config files? Are going to be supported on SF6+ by default?

ElGovanni:
Hi @markitosgv, as I see Fabian want to migrate everything to PHP even xml configs and I thinks is good to do it every where because we are getting IDE support.
https://github.com/symfony/symfony/issues/37186

markitosgv:
What about migrating [https://github.com/markitosgv/JWTRefreshTokenBundle/tree/master/Resources/config/mongodb](url) and [https://github.com/markitosgv/JWTRefreshTokenBundle/tree/master/Resources/config/orm](url) then?

ElGovanni:
Main purpose of this PR was to fix package for projects without [symfony/yaml](https://github.com/symfony/yaml) because for php/xml we don't need external parser, it would be nice if all configs were in php.

Issue #215, opened 2020-11-27
Undefined method 'findInvalid'. The method name must start with either findBy, findOneBy or countBy! 


ciklum-bohdan:
When I run 
```
bin/console gesdinet:jwt:clear -v 
```

I get an error

```

In EntityRepository.php line 235:
                                                                                                        
  [BadMethodCallException]                                                                              
  Undefined method 'findInvalid'. The method name must start with either findBy, findOneBy or countBy!  
                                                                                                        

Exception trace:
  at /projects/api-members/vendor/doctrine/orm/lib/Doctrine/ORM/EntityRepository.php:235
 Doctrine\ORM\EntityRepository->__call() at /projects/api-members/vendor/gesdinet/jwt-refresh-token-bundle/Doctrine/RefreshTokenManager.php:104
 Gesdinet\JWTRefreshTokenBundle\Doctrine\RefreshTokenManager->revokeAllInvalid() at /projects/api-members/vendor/gesdinet/jwt-refresh-token-bundle/Command/ClearInvalidRefreshTokensCommand.php:59
 Gesdinet\JWTRefreshTokenBundle\Command\ClearInvalidRefreshTokensCommand->execute() at /projects/api-members/vendor/symfony/console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /projects/api-members/vendor/symfony/console/Application.php:1018
 Symfony\Component\Console\Application->doRunCommand() at /projects/api-members/vendor/symfony/framework-bundle/Console/Application.php:97
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /projects/api-members/vendor/symfony/console/Application.php:271
 Symfony\Component\Console\Application->doRun() at /projects/api-members/vendor/symfony/framework-bundle/Console/Application.php:83
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /projects/api-members/vendor/symfony/console/Application.php:147
 Symfony\Component\Console\Application->run() at /projects/api-members/bin/console:42

gesdinet:jwt:clear [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<datetime>]

```

piciuok:
[...]

PR #233, opened 2021-03-05
RefreshTokenRepositoryInterface

piotrkreft:
Issue #215

The goal of the proposal is to provide a clear exception message for easier implementation of the solution.


piotrkreft:
Merged with #269

PR #241, opened 2021-04-23
Fixed compatibly with SF 5.3

lyrixx:


markitosgv:
Thank you so much @lyrixx

PR #244, opened 2021-06-10
Adapt RefreshTokenProvider for changes in Symfony 5.3

mbabker:
The `symfony/security-core` package has some changes that should be accounted for in the `RefreshTokenProvider` class:

- `UserProviderInterface` soft-adds a new `loadUserByIdentifier()` method, this will be required in Symfony 6
- `UserProviderInterface` deprecates the `loadUserByUsername()` method, this will be removed in Symfony 6
- The `Symfony\Component\Security\Core\User\User` class is deprecated in Symfony 5.3 and will be removed in Symfony 6, `Symfony\Component\Security\Core\User\InMemoryUser` should be used instead

The provider is updated to use the newer API when available and falls back to the legacy API in all other cases.


mbabker:
Technically, this requires #245 to land first because of the introduced typehints.

mbabker:
Incorporated into #246

PR #245, opened 2021-06-10
CI updates and bump minimum supported versions

mbabker:
Composer lists PHP 5.5 as the minimum supported version, yet this isn't actually tested.  So, I've bumped the PHP minimum up to 7.1.3 which matches Symfony 4 and the lowest tested PHP version.  This also helps unblock bringing in new code to address Symfony 5.3 deprecations.

I've also tightened the supported Symfony version list up so now only the 3.4 and 4.4 LTS' are supported, as well as only `^5.2` (this effectively drops 4.0-3 and 5.0-1).

The dependency list is updated to list out all of the actual dependencies instead of relying on transient dependencies to bring them in.

For CI, the matrix is updated to test PHP 7.4 and 8.0 (previously untested), to use Flex for restricting the Symfony version range (which avoids needing to install the `symfony/symfony` monorepo, which hid the broken dependency list), and makes sure all supported Symfony versions are tested.  There is also now a `--prefer-lowest` build in the queue to ensure the minimum dependencies are valid.

(On a side note, migrating to GitHub Actions would be a good idea here too)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,16 +13,29 @@
     "MIT"
   ],
   "require" : {
-    "php": "^5.5.9|>=7.0.8",
-    "symfony/framework-bundle": "~3.4|~4.0|~5.0",
-    "symfony/validator": "~3.4|~4.0|~5.0",
-    "lexik/jwt-authentication-bundle": "^1.1|^2.0@dev"
+    "php": ">=7.1.3",
+    "lexik/jwt-authentication-bundle": "^1.1|^2.0",
+    "symfony/config": "^3.4|^4.4|^5.2",
+    "symfony/console": "^3.4|^4.4|^5.2",
+    "symfony/dependency-injection": "^3.4.27|^4.4|^5.2",
+    "symfony/event-dispatcher": "^3.4|^4.4|^5.2",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -13,16 +13,29 @@
...
   "require" : {
-    "php": "^5.5.9|>=7.0.8",
-    "symfony/framework-bundle": "~3.4|~4.0|~5.0",
-    "symfony/validator": "~3.4|~4.0|~5.0",
-    "lexik/jwt-authentication-bundle": "^1.1|^2.0@dev"
+    "php": ">=7.1.3",
+    "lexik/jwt-authentication-bundle": "^1.1|^2.0",
+    "symfony/config": "^3.4|^4.4|^5.2",
+    "symfony/console": "^3.4|^4.4|^5.2",
+    "symfony/dependency-injection": "^3.4.27|^4.4|^5.2",
```
mbabker, line 20: The `spec\Gesdinet\JWTRefreshTokenBundle\DependencyInjection\GesdinetJWTRefreshTokenExtensionSpec` class makes use of the internal `ContainerBuilder::removeBindings()`, which was renamed at 3.4.27 from whatever it was previously, hence the minimum here being set on a later patch release.

Issue #242, opened 2021-05-21
Multiple authentication support


FabienLge:
Hello there ! I've been using this bundle since yesterday since it was recommended on the Lexik JWT doc, nethertheless I'm encountering problems with my setup that makes me simply unable to use it for now. I may have understood things wrong and maybe I'm not doing this correctly, you guys should be able to tell me !

For our own purposes, we have two distinct authentications, which are represented by two entites : **User** and **Customer**. Of course, there are a custom **UserProvider** for each of these entity. As I was following the documentation, I added the **UserProvider** as the configured User Provider as so : 
```yaml
user_provider: security.user.provider.concrete.app_user_provider
```

Then, I added an **EventSubscriber** subscribing to the **Kernel Response** event, in which I was refreshing the token like so : 
```php
if (($token = $this->tokenStorage->getToken()) instanceof JWTUserToken)
{
    /** @var User $user */
    $user = $token->getUser();

     $refreshTokenObject = $this->refreshTokenManager->getLastFromUsername($user->getEmail());
     $generatedRefreshToken = null;

     if (null !== $refreshTokenObject && $refreshTokenObject->isValid())
     {
        if ($refreshTokenObject->isValid())
        {
            $refreshToken = $refreshTokenObject->getRefreshToken();

            $req = Request::create("/token/refresh", "GET", ["refresh_token" => $refreshToken]);
                    
            /** @var JWTAuthenticationSuccessResponse $response */
            $response = $this->refreshService->refresh($req);

            $newTokens = json_decode($response->getContent(), true);

            $content = json_decode($event->getResponse()->getContent(), true);
                    
            $generatedRefreshToken = $newTokens["refresh_token"];
                    
            $responseContent = [
                "token" => $newTokens["token"],
                "response" => $content
             ];

             $event->setResponse($event->getResponse()->setContent(json_encode($responseContent)));
         }

         $this->refreshTokenManager->delete($refreshTokenObject);
     }
}
```

Here is the problem : since I was two user providers, if I connect myself on the one for **Customer**, my **Customer** instance will be converted into a **User** instance, which causes crashes as anyone would've expected.

Is there anything I can do about it ?

Thanks for the reply !

mbabker:
[...]

PR #246, opened 2021-06-10
New authenticator for the newer Symfony authenticator API

mbabker:
This is an implementation of an authenticator supporting the newer Symfony API.  It's very loosely based on a similar authenticator I wrote for one of our applications.  In theory, this should make setting up the firewall config a bit simpler, and allows deprecating the existing authentication path (namely `Gesdinet\JWTRefreshTokenBundle\Service\RefreshToken` and the existing `Gesdinet\JWTRefreshTokenBundle\Security\Authenticator\RefreshTokenAuthenticator`).

The intent is that a user can enable the authenticator for their refresh endpoint in a single spot, the firewall configuration:

```yaml
security:
    enable_authenticator_manager: true

    firewalls:
        api_token_refresh:
            pattern: ^/api/token/refresh
            stateless: true
            user_checker: App\Security\UserChecker
            refresh_jwt: ~
```

Which means the user doesn't have to define a controller for the refresh endpoint anymore (hence, the potential to deprecate `Gesdinet\JWTRefreshTokenBundle\Service\RefreshToken`).  This also means this bundle no longer has to manage the authentication flow on its own in a controller well after the Symfony authentication system has processed, it runs natively as a part of the core authenticator manager.

The authenticator is self-contained and goes through all the steps that the existing service and authenticator go through now regarding decoding the token, loading the token from storage and validating it, and refreshing the TTL when configured; all of this is now in a single class.  Because there isn't a "nice" way to pass along the decoded refresh token from the `authenticate()` method to the `onAuthenticationSuccess()` method, a custom authentication token is used as a middle layer for passing the refresh token through the authentication system without having to constantly re-query it.

This also decouples from the `LexikJWTAuthenticationBundle` failed authentication flow in favor of similar events and response handling native to this bundle, but better scoped to this bundle (so a downstream app can catch the new fail events from this bundle and handle that separately from a failure in the main JWT auth workflow).  The success handler is still used.

With the current implementation, this also helps in addressing #242 by identifying configuration that might need to be adjusted on a per-firewall basis.  Unfortunately, there are still some code paths (especially with the `LexikJWTAuthenticationBundle` authentication success handler) that need a more in-depth review/refactor to make this one work out well.

### TODO

- [x] Tests
- [x] Documentation
- [x] Deprecations


Event/RefreshEvent.php
```diff
@@ -12,27 +12,41 @@
...
     private $refreshToken;
 
-    private $preAuthenticatedToken;
+    /**
+     * @var TokenInterface
+     */
+    private $token;
 
-    public function __construct(RefreshTokenInterface $refreshToken, PostAuthenticationGuardToken $preAuthenticatedToken)
+    public function __construct(RefreshTokenInterface $refreshToken, TokenInterface $token)
```
mbabker, line 29: This is a type widening change, only the `Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken` token was accepted but with this PR the interface is allowed.

Event/RefreshEvent.php
```diff
@@ -12,27 +12,41 @@
...
 
     public function getRefreshToken()
     {
         return $this->refreshToken;
     }
 
+    /**
+     * @deprecated use getToken() instead
+     */
     public function getPreAuthenticatedToken()
```
mbabker, line 43: Deprecation since the token being retrieved here isn't guaranteed to be a pre-authentication, or post-authentication, token with the interface typehinted.

Resources/config/services.php
```diff
@@ -62,12 +63,23 @@
         ]);
 
     $services->set('gesdinet.jwtrefreshtoken.authenticator')
```
mbabker, line 65: This service, and the corresponding class, should be deprecated and removed when support for Symfony 5.2 and earlier is dropped.

markitosgv:
thanks @mbabker I need some time to review it

markitosgv:
@mbabker could you please resolve conflicts? thanks!

mbabker:
Rebased

mbabker:
... actually, looking at the diff, something fell out of sync between computers doing the rebase on this.  I'll have to fix it when I get home later.

mbabker:
OK, everything's good here.

PR #249, opened 2021-06-28
Switch to GitHub Actions

### LINKED ISSUE #248: migrate from travis to github actions
Migrate CI from travis-ci to github actions

mbabker:
Fixes #248

This is based on the config I use for various bundles supporting Symfony 3.4+.  

This also adds a job running on the in-development Symfony version, which is allowed to fail and can help to identify potential issues throughout the Symfony development cycle.


composer.json
```diff
@@ -54,5 +54,7 @@
     "branch-alias" : {
       "dev-master" : "1.0-dev"
     }
-  }
+  },
+  "minimum-stability": "dev",
```
maxhelias, line 58: This should not be added. If it blocks for testing, it is better to use the `composer config minimum-stability dev` command. Same for the prefer stable option

PR #250, opened 2021-07-01
Fix deprecation versions

mbabker:
Since a 1.0 beta was tagged, I'm guessing there won't be a 0.13.


markitosgv:
@mbabker you're right, the idea is release only security fixes on 0.12 and allow BC changes in 1.0.0

PR #256, opened 2021-07-13
Create a request extractor API to replace the static method for extracting the refresh token from the request

### LINKED ISSUE #95: Header based refreshToken
Hi all,

Please add another condition which checks the header of request for a refresh token:
```php
    public static function getRefreshToken(Request $request)
    {
        $refreshTokenString = null;
        if (false !== strpos($request->getContentType(), 'json')) {
            $content = $request->getContent();
            $params = !empty($content) ? json_decode($content, true) : array();
            $refreshTokenString = isset($params['refresh_token']) ? trim($params['refresh_token']) : null;
        } elseif (null !== $request->get('refresh_token')) {
            $refreshTokenString = $request->get('refresh_token');
        } elseif (null !== $request->request->get('refresh_token')) {
            $refreshTokenString = $request->request->get('refresh_token');
        } elseif (null !== $request->headers->get('authorization')) { // TODO: header check
            $refreshTokenString = $request->headers->get('authorization');
            $refreshTokenString = str_replace("Basic ", "", $refreshTokenString);
        }

        return $refreshTokenString;
    }
```
This will allow the users of [@juztcode/angular-auth](https://github.com/juztcode/angular-auth) use this bundle with Angular without pain.

mbabker:
Fixes #95

`RequestRefreshToken::getRefreshToken()` is a static method which makes it pretty difficult to customize the behavior for extracting the refresh token from the request data (as essentially you need to replace all services calling it with a custom implementation of those services then add a custom handler for pulling data from the request).  To get around this limitation, a new request extractor API (inspired by the similar API in `LexikJWTAuthenticationBundle`) is introduced with this PR and the current static method removed.

The default implementation that is aliased to the new interface is a chain extractor that will iterate over all extractors until finding a token.  Similar to the current behavior, null is returned from the extractor API if a token can't be found.

Extractors can be added in applications by implementing `Gesdinet\JWTRefreshTokenBundle\Request\Extractor\ExtractorInterface` (which is configured to support autowiring and automatically tagging the class with the `gesdinet_jwt_refresh_token.request_extractor` container tag).  The tag also supports priorities, so applications can prioritize the extractors as desired.

```yaml
services:
    gesdinet.jwtrefreshtoken.request.extractor.request_body:
        class: Gesdinet\JWTRefreshTokenBundle\Request\Extractor\RequestBodyExtractor
        tags:
            - { name: gesdinet_jwt_refresh_token.request_extractor, priority: 25 }
```


markitosgv:
@mbabker What do you think about remove instead of marking as deprecated because we're going to release it in a BC major update (v1.0) , maybe we need to introduce a CHANGELOG.md

mbabker:
Changelog file started, removed the old class, and updated the README.

Resources/config/services.php
```diff
@@ -54,6 +59,22 @@
...
 
+    $services->set('gesdinet.jwtrefreshtoken.request.extractor.chain')
+        ->class(ChainExtractor::class)
+        ->public();
+
+    $services->alias(ExtractorInterface::class, 'gesdinet.jwtrefreshtoken.request.extractor.chain');
+
+    $services->set('gesdinet.jwtrefreshtoken.request.extractor.request_body')
+        ->class(RequestBodyExtractor::class)
+        ->public()
```
maxhelias, line 70: I think we don't need to have the extractors in public, the chain is enough.

mbabker: Yeah, you’re probably right.  That was most likely just excessive copy/paste.

maxhelias: It's just a detail, don't listen me 😄 
It's a great evolution in any case, good job !

PR #258, opened 2021-07-15
Test improvements

mbabker:
Puts in some effort to improve the bundle's test coverage.

The biggest change is that the repository classes now use functional tests against a real database instance.  This also helps with making sure the schema mappings are valid (which, they weren't for MongoDB).  Beyond that, it's mostly filling in specs around some of the newer classes or covering the non-default and error-path behaviors in some classes.

This also bumps up some of the development dependencies to more recent minimums.  Some of the minimums actually error out on the prefer minimum build for PHP 7.4, so these should be set to something that doesn't cause errors.


markitosgv:
Great work @mbabker !!

PR #199, opened 2020-08-10
Support for refresh token in cookie

lukacovicadam:
Hi @markitosgv 

I created support for refresh token in cookie. Cookie is also automatically extracted. Cookie is by default httpOnly, thus cannot be read by JS and it's safe against XSS attack.

PR #274, opened 2021-08-14
Support setting refresh token in cookie

Jayfrown:
This commit is the result of cherry-picking the following commits on top
of the current master branch, resolving conflicts, making some minor
changes and rebasing the lot into a single commit:

https://github.com/markitosgv/JWTRefreshTokenBundle/pull/199/commits/c6f67b097e7d80070464d3150459602c5f3b18e0
https://github.com/markitosgv/JWTRefreshTokenBundle/pull/199/commits/f9fbdf56e36fb7d64a6ac03516995a3448e34096
https://github.com/markitosgv/JWTRefreshTokenBundle/pull/199/commits/3267470c954c6de1000612967eb96cdf56325768

Resolved conflicts in:
	DependencyInjection/Configuration.php
	DependencyInjection/GesdinetJWTRefreshTokenExtension.php
	EventListener/AttachRefreshTokenOnSuccessListener.php
	README.md
	Request/RequestRefreshToken.php
	Resources/config/services.php
	spec/EventListener/AttachRefreshTokenOnSuccessListenerSpec.php
	spec/Request/RequestRefreshTokenSpec.php


DependencyInjection/Configuration.php
```diff
@@ -84,6 +84,19 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->info('When true, resolving of Doctrine mapping is done automatically to use either ORM or ODM object manager')
                     ->defaultTrue()
                 ->end()
+                ->scalarNode('token_parameter_name')->defaultValue('refresh_token')->end()
```
mbabker, line 87: `token_parameter_name` is duplicated now (it's set a few lines up)

DependencyInjection/Configuration.php
```diff
@@ -84,6 +84,19 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                ->arrayNode('cookie')
```
mbabker, line 88: IMO this probably needs a `canBeEnabled()` call which would result in defaulting the cookie storage to disabled.

Reasons for this:

1) The checks in the event listener right now boil down to whether or not the cookie storage settings array is empty, which if the config defaults are being set, would never be the case
2) Enabling cookie storage by default would create a rather interesting B/C break, the cookie-based approach really only works well when you're using an API behind a SPA but in more traditional REST API applications having the token be in a cookie might be problematic.

DependencyInjection/GesdinetJWTRefreshTokenExtension.php
```diff
@@ -46,6 +46,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('gesdinet_jwt_refresh_token.single_use', $config['single_use']);
         $container->setParameter('gesdinet_jwt_refresh_token.token_parameter_name', $config['token_parameter_name']);
         $container->setParameter('gesdinet_jwt_refresh_token.doctrine_mappings', $config['doctrine_mappings']);
+        $container->setParameter('gesdinet_jwt_refresh_token.cookie', (isset($config['cookie']) ? $config['cookie'] : []));
```
mbabker, line 49:
```suggestion
        $container->setParameter('gesdinet_jwt_refresh_token.cookie', $config['cookie'] ?? []);
```

PHP minimum bumped up since the original PRs, so optimize the syntax as well.

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -68,7 +74,8 @@ public function __construct(
         $tokenParameterName,
         $singleUse,
         RefreshTokenGeneratorInterface $refreshTokenGenerator,
-        ExtractorInterface $extractor
+        ExtractorInterface $extractor,
+        $cookie
```
mbabker, line 78:
```suggestion
        array $cookieSettings
```

Typehint the parameter and rename the parameter to make it a bit more clear what this array actually is (it's not an array with the cookie data within it, it's an array with the settings to configure a response cookie).

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -77,6 +84,7 @@ public function __construct(
         $this->singleUse = $singleUse;
         $this->refreshTokenGenerator = $refreshTokenGenerator;
         $this->extractor = $extractor;
+        $this->cookie = $cookie;
```
mbabker, line 87: Maybe this should have a default structure set (similar to how the 5.3 authenticator sets defaults for its config array in the constructor)?

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -107,9 +115,27 @@ public function attachRefreshToken(AuthenticationSuccessEvent $event): void
             $refreshToken = $this->refreshTokenGenerator->createForUserWithTtl($user, $this->ttl);
 
             $this->refreshTokenManager->save($refreshToken);
-            $data[$this->tokenParameterName] = $refreshToken->getRefreshToken();
+            $refreshTokenString = $refreshToken->getRefreshToken();
         }
 
-        $event->setData($data);
+        if ($this->cookie) {
```
mbabker, line 121: If the suggestion about adding `canBeEnabled()` in the config is applied, then this should be changed.  Otherwise, as pointed out above, right now this enables itself based on whether the config array is empty or not.

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -107,9 +115,27 @@ public function attachRefreshToken(AuthenticationSuccessEvent $event): void
...
+            $event->getResponse()->headers->setCookie(
+                new Cookie(
+                    $this->tokenParameterName,
+                    $refreshTokenString,
+                    time() + $this->ttl,
+                    $this->cookie['path'],
```
mbabker, line 127: Since there aren't any defaults set right now, either defaults need to be set (as suggested above) or checks to make sure all of these keys are defined need to be added with appropriate fallbacks.

Jayfrown:
Thanks for reviewing @mbabker, I believe I've fixed the things you mentioned.

Jayfrown:
Hey @markitosgv have you had a chance to look at this? Let me know if you would like to see some changes or if you have any other reservations

DependencyInjection/Configuration.php
```diff
@@ -84,6 +84,19 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->info('When true, resolving of Doctrine mapping is done automatically to use either ORM or ODM object manager')
                     ->defaultTrue()
                 ->end()
+                ->arrayNode('cookie')
+                    ->canBeEnabled()
+                    ->children()
+                        ->enumNode('sameSite')
```
chalasr, line 90: I suggest using snake_case for option names for consistency

Jayfrown: Thanks @chalasr - I've updated the PR.

maxhelias (approved):

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -107,9 +121,27 @@ public function attachRefreshToken(AuthenticationSuccessEvent $event): void
...
+                    time() + $this->ttl,
+                    $this->cookieSettings['path'],
+                    $this->cookieSettings['domain'],
+                    $this->cookieSettings['secure'],
+                    $this->cookieSettings['http_only'],
+                    false,
+                    $this->cookieSettings['same_site']
+                )
+            );
+        } else {
```
djpremier, line 141: Could we put an "if" as an argument to allow us to use both? How to raise this issue

https://github.com/lexik/LexikJWTAuthenticationBundle/issues/909

djpremier, line 141: Could we put an "if" as an argument to allow us to use both? How to raise this issue

https://github.com/lexik/LexikJWTAuthenticationBundle/issues/909

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -107,9 +121,27 @@ public function attachRefreshToken(AuthenticationSuccessEvent $event): void
...
+            $data[$this->tokenParameterName] = $refreshTokenString;
+
+            $event->setData($data);
```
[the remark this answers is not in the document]
Jayfrown: If I understand correctly you would like to be able to use both `refresh_token` in response cookie as well as the `JsonResponse`?

In my opinion, since obtaining a `refresh_token` is enough to impersonate a user and gain indefinite access to the application, it is paramount to ensure the `refresh_token` is not stolen. By setting it in the `JsonResponse`, JavaScript executed in the browser will be able to read the `refresh_token` from the response, while JavaScript would not be able to read the value from a HttpOnly cookie.

I think, if cookies are enabled, by default it should not send the `refresh_token` in the `JsonResponse` to protect from this scenario.

I will have a look if I can do something similar to https://github.com/lexik/LexikJWTAuthenticationBundle/issues/909 to allow users to explicitly enable this behavior if it is really needed.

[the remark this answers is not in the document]
Jayfrown: @djpremier Thanks for the suggestion, [this](https://github.com/markitosgv/JWTRefreshTokenBundle/pull/274/commits/d4bd2b1f9428c593d49b05d561006f40a11a4204) should do what you asked for.

```
gesdinet_jwt_refresh_token:
    cookie:
      enabled: true
      remove_token_from_body: false
```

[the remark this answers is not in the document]
djpremier: Hi @Jayfrown, yes I also agree that the best is to use the cookie, but as raised in the issue, if the application is geared towards a globally-used API (via command line for example), or in the case of a mobile application, have access to the refresh_token is required.

Thank you very much for adding the functionality, the fact that it is optional is the best, because if you ever need to use it, it's simple.

chalasr (approved):

CIT-SW:
Any news when this will be merged?

CIT-SW:
@markitosgv 
Hey there! We completely understand when an open source project can't get the amount of attention from the author that would be needed and we really appreciate you're work. Nevertheless it's a bit hard to figure out if we could actually use your package or not in our project as long as we can't tell if and when refresh token via cookie will be available (and if the package is going to be further maintained at a minimum level).
Hence, would you be so kind as to let us know if and when this PR can be merged? Of course, "no" is a valid answer as well, it would even be better to know that you are not going to merge this PR / maintain the package than no knowing anything about it. ;-)
Looking forward to your reply!

markitosgv:
checked and merged ;-)

PR #284, opened 2021-11-06
Drop support for Symfony 3.4 and 5.3, add support for Symfony 6, drop MongoDB ODM 1.x

mbabker:
Symfony 3.4 and MongoDB ODM 1.x are EOL, there isn't much of a point in keeping support for them now (one can use older versions if they still need it).

Symfony 5.3 goes EOL this month (January 2022), let's drop it now and simplify a bit of the new authenticator API code.

Symfony 6 is here, this adds the support needed for it.


Doctrine/RefreshTokenManager.php
```diff
@@ -41,12 +41,15 @@ class RefreshTokenManager implements RefreshTokenManagerInterface
     public function __construct(ObjectManager $om, $class)
     {
         $this->objectManager = $om;
-        $this->repository = $om->getRepository($class);
 
-        if (!$this->repository instanceof RefreshTokenRepositoryInterface) {
+        $repository = $om->getRepository($class);
+
+        if (!$repository instanceof RefreshTokenRepositoryInterface) {
```
mbabker, line 47: I made this tweak after doing a PHPStan analysis, this avoids setting the class property if the type is incorrect (if the property would typed, this would result in a type error instead of the exception below)

Event/RefreshAuthenticationFailureEvent.php
```diff
@@ -13,14 +13,15 @@
...
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Contracts\EventDispatcher\Event;
 
 class RefreshAuthenticationFailureEvent extends Event
 {
     private AuthenticationException $exception;
 
-    private ?Response $response;
+    private Response $response;
```
mbabker, line 22: SF6 typehints the response as non-nullable in authentication success and failure handlers, so the events shouldn't allow nulllability either

Http/RefreshAuthenticationFailureResponse.php
```diff
@@ -14,7 +14,41 @@
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Response;
 
-class RefreshAuthenticationFailureResponse extends JsonResponse
+if (80000 <= \PHP_VERSION_ID && (new \ReflectionMethod(JsonResponse::class, 'setData'))->hasReturnType()) {
```
mbabker, line 17: B/C layer because in SF6 there are PHP 8 types used on the class, so the eval hack is needed for PHP 7.4 compat.

Model/AbstractRefreshToken.php
```diff
@@ -56,7 +56,7 @@ public static function createForUserWithTtl(string $refreshToken, UserInterface
      */
     public function __toString()
     {
-        return $this->getRefreshToken();
+        return $this->getRefreshToken() ?: '';
```
mbabker, line 59: `getRefreshToken()` has a nullable return, `__toString()` should always be a string.

README.md
```diff
@@ -25,6 +25,8 @@ If you want to use this bundle with previous Symfony versions, please use 0.2.x
 
 **It's important you manually require either Doctrine's ORM or MongoDB ODM as well, these packages are not required automatically as you can choose between them. Failing to do so may trigger errors on installation**
 
+If using Symfony 5.2 or older, you will also need to install the `symfony/security-guard` package, it is only required for the legacy authentication API and is not compatible with Symfony 6.0.
```
mbabker, line 28: The Guard component is deprecated and dropped in Symfony 6.  I made it an optional dependency, folks on Symfony 4.4 will need to manually install that component to keep using this package.  Folks on Symfony 5.3 and later won't need to have it installed since the newer API doesn't use it.

Security/Provider/RefreshTokenProvider.php
```diff
@@ -129,6 +129,6 @@ public function supportsClass($class)
             return true;
         }
 
-        return User::class === $class;
+        return class_exists(User::class) && User::class === $class;
```
mbabker, line 132: SF6 drops the User class, so technically the provider shouldn't support the class if it's not available.

composer.json
```diff
@@ -15,22 +15,21 @@
...
-    "symfony/security-guard": "^3.4|^4.4|^5.2",
-    "symfony/security-http": "^3.4|^4.4|^5.2"
+    "symfony/property-access": "^4.4|^5.3|^6.0",
+    "symfony/security-bundle": "^4.4|^5.3|^6.0",
+    "symfony/security-core": "^4.4|^5.3|^6.0",
+    "symfony/security-http": "^4.4|^5.3|^6.0"
   },
   "require-dev": {
-    "akeneo/phpspec-skip-example-extension": "^4.0|^5.0",
+    "babdev/phpspec-skip-example-extension": "^1.0",
```
mbabker, line 32: The source package for this looks to be not well maintained and doesn't actually support skipping examples with requirements declared at the example level.  So, this changes to my fork that adds the feature.

DependencyInjection/GesdinetJWTRefreshTokenExtension.php
```diff
@@ -51,7 +52,8 @@ public function load(array $configs, ContainerBuilder $container): void
         $refreshTokenClass = RefreshTokenEntity::class;
         $objectManager = 'doctrine.orm.entity_manager';
 
-        if (!class_exists(DoctrineOrmMappingsPass::class) || 'mongodb' === strtolower($config['manager_type'])) {
+        // Change the refresh token and object manager to the MongoDB ODM if the configuration explicitly sets it or if the ORM is not installed and the MongoDB ODM is
+        if ('mongodb' === strtolower($config['manager_type']) || (!class_exists(EntityManager::class) && class_exists(DocumentManager::class))) {
```
mbabker, line 56: Changing this to use the object managers instead of the compiler passes in the bundles allows to break the dev dependencies on the bundles (which, were only pulled in for these checks).

Service/RefreshToken.php
```diff
@@ -85,18 +84,16 @@ public function __construct(
...
+        $credentials = $this->authenticator->getCredentials($request);
+
         try {
-            $user = $this->authenticator->getUser(
-                $this->authenticator->getCredentials($request),
-                $this->provider
-            );
+            $user = $this->authenticator->getUser($credentials, $this->provider);
 
             $postAuthenticationToken = $this->authenticator->createAuthenticatedToken($user, $this->providerKey);
```
bravik, line 92: this should probably removed or deprecated too, as `PostAuthenticationGuardToken` is deprecated.
For `->onAuthenticationSuccess()` any `TokenInterface` will work, as it only uses `getUser()`.
And for `RefreshToken` event getting pre authenticated toeken should probably be deprecated or reworked.. I'm not sure why is it needed at all

ad3n:
Any update?

mbabker:
Blocked by phpspec not supporting Symfony 6 yet.

markitosgv:
merged @webignition #293

composer.json
webignition: Thankfully now not needed :)

mbabker:
OK, we're good to go here.

README.md
```diff
@@ -130,7 +132,7 @@ security:
 # ...
 ```
 
-### Step 3 (Symfony 5.2-)
+### Step 3 (Symfony 4.4)
```
webignition, line 135: A little below this point there is still a reference to Symfony 3:

```
**Symfony 3 Version:**
```yaml
# app/config/routing.yml
gesdinet_jwt_refresh_token:
    path:     /api/token/refresh
    defaults: { _controller: gesdinet.jwtrefreshtoken:refresh }
# ...
```

mbabker: I've done a once over on the README.

PR #286, opened 2021-11-17
Update README.md

sukhoy94:
Need to add this, anyway there is an error: 
"Class XXX is not a valid entity or mapped super class"


sukhoy94:
still not working, sorry

Issue #255, opened 2021-07-12
Testing before 1.0.0 release 


markitosgv:
Hi!

In a few time we are going to release a v1.0 stable version, now we are testing v1.0.0-beta2 . Could you please give me some feedback to try to fix some issues before releasing it? Thanks!!

pmikolajek:
I'm on `v1.0.0-beta2`, and I'm getting some deprecation notices when running symfony commands like `bin/console list` or `bin/console doctrine:schema:validate`:

```
2021-07-15T22:34:14+00:00 [info] User Deprecated: Since gesdinet/jwt-refresh-token-bundle 1.0: The "Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManager" class is deprecated, implement "Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManagerInterface" directly.

2021-07-15T22:29:24+00:00 [info] User Deprecated: The "Gesdinet\JWTRefreshTokenBundle\Model\AbstractRefreshToken::setValid()" method will require a new "\DateTimeInterface|null $refreshToken" argument in the next major version of its interface "Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenInterface", not defining it is deprecated.

2021-07-15T22:29:24+00:00 [info] User Deprecated: The "Gesdinet\JWTRefreshTokenBundle\Model\AbstractRefreshToken::setUsername()" method will require a new "string|null $refreshToken" argument in the next major version of its interface "Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenInterface", not defining it is deprecated.
```
 
- The first one seems to be because `\Gesdinet\JWTRefreshTokenBundle\Doctrine\RefreshTokenManager` uses the deprecated `Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManager`.
- The latter two are caused by what look like invalid phpdocs for `setValid()` and `setUsername()` methods in `\Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenInterface`.

I'm on a mostly-default setup from docs. Am I doing something wrong?

mbabker:
The doc blocks should be fixed up after https://github.com/markitosgv/JWTRefreshTokenBundle/pull/259

[...]

PR #299, opened 2022-01-20
Add firewall example for multiple user providers

webignition:
Regarding #255 and the need to specify the `provider` key if used within an app with more than one user provider, here is a minimal update to the readme with commented-out examples of how to specify the required user provider.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


webignition:
@shakaran Yep, doing so now. Whilst github's fork syncing option I removed my commit which then caused this PR to be auto-closed. Adding my changes again now ...

webignition:
Well seems like this PR is well and truly accidentally closed :smile: 

New PR (with same changes and no conflicts) at #352

PR #302, opened 2022-02-05
Fixes #287 - Add LogoutEventListener

### LINKED ISSUE #287: UX with cookie-based refresh tokens
Hello there! 👋 

I've been testing the "cookie" option of this bundle, lately. 
This is awesome and it works great! 🎉  It avoids having to store the token on the client side and expose it to creepy people.

However, I notice 2 usability issues with this:
1. When the user logs out from the SPA, client just destroys the access token and the user gets redirected to the login page. Fine. problem is, since client has not access to the refreshToken anymore, it can no longer destroy it. This means that next call to the refresh token route will log the user in, even if they wanted to logout, because the cookie is still here 😓 

2. Since the refreshToken is no longer the client's business, 2 different users can't be logged in within the same browser. I usually have an "admin" tab and a "customer" tab. Since both share the same cookie (same API), as soon as I sign in as a customer, I get kicked from the admin tab once this one asks for a new token.

Regarding **1**: how about the following flow:
```http
# Refresh current token
POST /api/token/refresh
Cookie: refreshToken=myCurrentRefreshToken

HTTP/2 200 OK
Set-Cookie: refreshToken=someOtherRefreshToken, expires=in one week

# Revoke that token
DELETE /api/token/refresh
Cookie: refreshToken=someOtherRefreshToken

HTTP/2 200 OK
Set-Cookie: refreshToken=, expires=right now
```

Regarding **2**... I have no real idea on how to address this. Thoughts?

Jayfrown:
This commit introduces a `LogoutEventListener` which invalidates the given `refresh_token` and unsets the cookie, if enabled.

If no refresh token is supplied, an error is returned and the cookie remains untouched. If the supplied refresh token is (already) invalid, the cookie is unset.

Because the `LogoutEventListener` always sets a response, it would inhibit normal logout behavior and therefore should only run on a specifically configured firewall.

Therefore a new configuration option is introduced, called `logout_firewall`, which contains the name of the firewall that triggers the logout event we want to hook into (default: `api`).

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -98,6 +98,10 @@ class Configuration implements ConfigurationInterface
                         ->scalarNode('remove_token_from_body')->defaultTrue()->end()
                     ->end()
                 ->end()
+                ->scalarNode('logout_firewall')
+                    ->defaultValue('api')
+                    ->info('Name of the firewall that triggers the logout event to hook into (default: api)')
+                ->end()
             ->end();
 
         return $treeBuilder;
--- a/EventListener/LogoutEventListener.php
+++ b/EventListener/LogoutEventListener.php
@@ -1,0 +1,109 @@
+<?php
+
+/*
+ * This file is part of the GesdinetJWTRefreshTokenBundle package.
+ *
+ * (c) Gesdinet <http://www.gesdinet.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gesdinet\JWTRefreshTokenBundle\EventListener;
+
+use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManagerInterface;
+use Gesdinet\JWTRefreshTokenBundle\Request\Extractor\ExtractorInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Security\Http\Event\LogoutEvent;
+
+class LogoutEventListener
+{
+    private RefreshTokenManagerInterface $refreshTokenManager;
+    private ExtractorInterface $refreshTokenExtractor;
+    private string $tokenParameterName;
+    private array $cookieSettings;
+    private string $logout_firewall_context;
+
+    public function __construct(
+        RefreshTokenManagerInterface $refreshTokenManager,
+        ExtractorInterface $refreshTokenExtractor,
+        string $tokenParameterName,
+        array $cookieSettings,
+        string $logout_firewall_context,
+    ) {
+        $this->refreshTokenManager = $refreshTokenManager;
+        $this->refreshTokenExtractor = $refreshTokenExtractor;
+        $this->tokenParameterName = $tokenParameterName;
+        $this->cookieSettings = array_merge([
+            'enabled' => false,
+            'same_site' => 'lax',
+            'path' => '/',
+            'domain' => null,
+            'http_only' => true,
+            'secure' => true,
+            'remove_token_from_body' => true,
+        ], $cookieSettings);
+        $this->logout_firewall_context = $logout_firewall_context;
+    }
+
+    public function onLogout(LogoutEvent $event): void
+    {
+        $request = $event->getRequest();
+        $current_firewall_context = $request->attributes->get('_firewall_context');
+
+        if ($current_firewall_context !== $this->logout_firewall_context) {
+            return;
+        }
+
+        $tokenString = $this->refreshTokenExtractor->getRefreshToken($request, $this->tokenParameterName);
+        if (null === $tokenString) {
+            $event->setResponse(
+                new JsonResponse(
+                    [
+                        'code' => 400,
+                        'message' => 'No refresh_token found.',
+                    ],
+                    JsonResponse::HTTP_BAD_REQUEST
+                )
+            );
+
+            return;
+        } else {
+            $refreshToken = $this->refreshTokenManager->get($tokenString);
+            if (null === $refreshToken) {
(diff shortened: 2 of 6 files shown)


Resources/config/services.php
```diff
@@ -157,4 +158,17 @@
...
+
+    $services->set(LogoutEventListener::class)
+        ->args([
+            new Reference('gesdinet.jwtrefreshtoken.refresh_token_manager'),
+            new Reference('gesdinet.jwtrefreshtoken.request.extractor.chain'),
+            new Parameter('gesdinet_jwt_refresh_token.token_parameter_name'),
+            new Parameter('gesdinet_jwt_refresh_token.cookie'),
+            new Parameter('gesdinet_jwt_refresh_token.logout_firewall_context'),
+        ])
+        ->tag('kernel.event_listener', [
```
Jayfrown, line 170: Would've preferred to set `'dispatcher' => 'security.event_dispatcher.{$logout_firewall}'` to avoid calling the `LogoutEventListener` altogether instead of checking the firewall context within the listener, however I couldn't figure out how to reference a parameter value from this scope.

Jayfrown:
@mbabker If you have the time, would you mind reviewing this PR? Maybe we can spot and fix some issues before @markitosgv has the time to take a look at it.

DependencyInjection/Configuration.php
```diff
@@ -98,6 +98,9 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->scalarNode('remove_token_from_body')->defaultTrue()->end()
                     ->end()
                 ->end()
+                ->scalarNode('logout_firewall')
+                    ->defaultValue('api')
+                    ->info('Name of the firewall that triggers the logout event to hook into (default: api)')
```
mbabker, line 103:
```suggestion
                    ->info('Name of the firewall that triggers the logout event to hook into (default: api)')
                ->end()
```

Looks like an `end()` call is missing based on the indents.

EventListener/LogoutEventListener.php
```diff
@@ -0,0 +1,126 @@
...
+
+        $tokenString = $this->refreshTokenExtractor->getRefreshToken($request, $this->tokenParameterName);
+        if (null === $tokenString) {
+            $event->setResponse(
+                new JsonResponse(
+                    [
+                        'code' => 422,
+                        'message' => 'No refresh_token found.',
+                    ],
+                    JsonResponse::HTTP_UNPROCESSABLE_ENTITY
```
mbabker, line 85: Do we really need to have logout trigger a 422 response if the refresh token isn't part of the request?

Resources/config/services.php
mbabker: The other option I can think of is to register the service in the container extension class, that way you have the bundle config available.

But, I don't think it's worth the effort at the moment.  The bundle already has a bunch of other quirks that make it difficult to use in a multi-firewall setup (Sylius is probably the most practical example I can think of where it has multiple firewalls, you wouldn't be able to handle refresh tokens for the shop and admin API contexts separately).  So it's fine for now to go with this and later on if things are reworked then the event listener could be attached to the right dispatcher(s).

EventListener/LogoutEventListener.php
```diff
@@ -0,0 +1,126 @@
...
+use Gesdinet\JWTRefreshTokenBundle\Request\Extractor\ExtractorInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Security\Http\Event\LogoutEvent;
+
+class LogoutEventListener
+{
+    /**
+     * @var RefreshTokenManagerInterface
+     */
+    private $refreshTokenManager;
```
mbabker, line 24: We can use native typehints for new classes with the PHP 7.4 minimum.

DependencyInjection/Configuration.php
Jayfrown: Thanks, fixed.

EventListener/LogoutEventListener.php
Jayfrown: I'm on the fence. If we don't, the cookie might be cleared on the client-side but the refresh token will not be invalidated on the API side.

mbabker: So early return in the case where there is no refresh token in the request instead of sending a 422 which also has the side effect of invalidating the client cookie, and then only invalidate the cookie if this listener is going to set a 200 response.  Most 4xx responses indicate the client can change something about the request and retry it, but if you invalidate the cookie even when there's no token in the request, then there isn't a way to retry the request.  I think a 400 response in this case might also be more appropriate than 422.

EventListener/LogoutEventListener.php
```diff
@@ -0,0 +1,107 @@
...
+            if (null === $refreshToken) {
+                $event->setResponse(
+                    new JsonResponse(
+                        [
+                            'code' => 422,
+                            'message' => 'Invalid refresh_token supplied.',
+                        ],
+                        JsonResponse::HTTP_UNPROCESSABLE_ENTITY
+                    )
+                );
```
mbabker, line 80: Now that I think about it more, I wonder if a 4xx response is even appropriate in this case.  If you're trying to invalidate a token that doesn't exist in the system, the desired end state (invalidated token) is still reached.  It's not much different than ignoring errors if the file you're trying to delete from the filesystem already doesn't exist.

```php
try {
    $this->filesystem->delete($file);
} catch (FileNotFoundException) {
    // If it's gone then we're OK
}
```

Jayfrown: Makes sense. I changed this to send a 200 response indicating the supplied refresh token is already invalid. It still falls through and unsets the cookie.

PR #303, opened 2022-02-06
Check the request path in the authenticator

mbabker:
Related to https://github.com/markitosgv/JWTRefreshTokenBundle/issues/255#issuecomment-1030286770 and some other comments and issues

This should improve the DX and the use of the authenticator with Symfony 5.4+ by making the authenticator check the request path instead of relying on the presence of the refresh token in the request.  Relying on the token being in the request to trigger the authenticator is deprecated with this PR and will be removed in 2.0 in favor of only matching the request path.

A nice side effect of this change is that the security configuration becomes much cleaner.  Using my tester app, I was able to simplify things down with [this commit](https://github.com/mbabker/refresh-token-tester/commit/3deb2497f7d8bfcaea4a4071a6cbae0aeb7b1c8f) and now all of my API related authentication is on one firewall instead of three.

### DIFF
--- a/Security/Http/Authenticator/RefreshTokenAuthenticator.php
+++ b/Security/Http/Authenticator/RefreshTokenAuthenticator.php
@@ -35,6 +35,7 @@ use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
 use Symfony\Component\Security\Http\Authenticator\Passport\SelfValidatingPassport;
 use Symfony\Component\Security\Http\Authenticator\Passport\UserPassportInterface;
 use Symfony\Component\Security\Http\EntryPoint\AuthenticationEntryPointInterface;
+use Symfony\Component\Security\Http\HttpUtils;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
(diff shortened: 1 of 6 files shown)


Security/Http/Authenticator/RefreshTokenAuthenticator.php
```diff
@@ -69,14 +73,26 @@ public function __construct(
         $this->successHandler = $successHandler;
         $this->failureHandler = $failureHandler;
         $this->options = array_merge([
+            'check_path' => null, // @todo in 2.0, change the default to `/login_check`
```
Jayfrown, line 76: > todo in 2.0, change the default to `/login_check`

This refers to the `check_path` configuration option on the `refresh_jwt` authenticator, right? Why default to `/login_check`, shouldn't it be something related to refresh rather than login?

Jayfrown (approved):

Security/Http/Authenticator/RefreshTokenAuthenticator.php
mbabker: I just took the same path Symfony uses in all of the authenticators for their defaults, this could be anything really.

maxhelias (approved):
Very good improvement 👌

PR #304, opened 2022-02-06
[B/C Break] Change the object mappings to mapped superclasses

### LINKED ISSUE #285: Use another table for refresh tokens
Is it possible to use another table name to store and check refresh tokens instead of public.refresh_tokens ?

mbabker:
Fixes #285

This fixes the inability to customize the storage schema for the refresh token object (i.e. anything where you try to extend the provided Document or Entity classes to do things like change the table name or add fields).  Unfortunately, this does have a practical B/C break because when downstream users pull in the updated version with this change, they will need to make some changes in their application (specifically the bits noted in the updated README about configuring the bundle from the Flex recipe).

If accepted, I can get https://github.com/symfony/recipes-contrib/compare/master...mbabker:jwt-refresh-bundle submitted to add the needed recipe to the Flex contrib repo, which will also help new installs or anyone updating their app (and also allowing contrib recipes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- [B/C Break] Changed the object mappings to mapped superclasses, this requires updating your app's configuration
 - Added support for checking the request path in the `refresh_jwt` authenticator
 - Deprecated not configuring the request path to check in the `refresh_jwt` authenticator
 
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -52,6 +52,16 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('refresh_token_class')
                     ->defaultNull()
                     ->info(sprintf('Set the refresh token class to use (default: %s)', RefreshToken::class))
+                    ->validate()
+                        ->ifTrue(static fn ($v): bool => null === $v)
+                        ->then(static function () {
+                            trigger_deprecation(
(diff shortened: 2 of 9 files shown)


DmytryHo:
Is the any plan for this fix to be released?
Symfony 6 support added only in 1.0.0 version, but update to 1.0.0 still break ability to use custom table for tokens.

mboron83:
> Is the any plan for this fix to be released? Symfony 6 support added only in 1.0.0 version, but update to 1.0.0 still break ability to use custom table for tokens.

This is exactly what I'd like to ask. This feature is extremely needed :-)

ricardodevries:
@markitosgv Is there any projection on when this pull request can be merged followed by a new release?

markitosgv:
@mbabker do you think is needed to release a v2 version?

mbabker:
> do you think is needed to release a v2 version?

By SemVer, I don't think this PR strictly has to force a 2.0 release as there aren't any API changes involved.  But, there is a practical break for every consuming application because they'll have to add an empty entity and a line of configuration when upgrading.

Also given that 1.0 is broken in a few cases (sorry about that), it might be better to just tag it as 1.1 given it's fixing some broken workflows and features that were lost from the 0.12 release.

All told, I actually don't know what the right answer here is.  If 2.0 is decided though, I'd say anything marked deprecated in the codebase right now should have those notices updated to reflect removal in 3.0 instead.

mbabker:
For anyone interested, https://github.com/symfony/recipes-contrib/pull/1390 is the Flex recipe PR.

PR #309, opened 2022-04-07
Expiration ttl parameter

ricardodevries:
This adds a expiration timestamp to the response, if enabled. This feature is disabled by default.

> ### Return Expiration Timestamp
> 
> If set to true, the expiration timestamp will be added to the response.
> 
> ```yaml
> gesdinet_jwt_refresh_token:
>     return_expiration: true
> ```
> 
> The default parameter name is `refresh_token_expiration`. You can change the parameter name by adding this line to your config and changing it:
> 
> ```yaml
> gesdinet_jwt_refresh_token:
>     return_expiration_parameter_name: refresh_token_expiration
> ```

### DIFF
--- a/EventListener/AttachRefreshTokenOnSuccessListener.php
+++ b/EventListener/AttachRefreshTokenOnSuccessListener.php
@@ -60,9 +60,21 @@ class AttachRefreshTokenOnSuccessListener
     protected array $cookieSettings;
 
     /**
+     * @var bool
+     */
+    protected $returnExpiration;
+
+    /**
+     * @var string
+     */
+    protected $returnExpirationParameterName;
+
+    /**
      * @param int    $ttl
      * @param string $tokenParameterName
      * @param bool   $singleUse
+     * @param bool   $returnExpiration
+     * @param string $returnExpirationParameterName
      */
     public function __construct(
         RefreshTokenManagerInterface $refreshTokenManager,
@@ -72,7 +84,9 @@ class AttachRefreshTokenOnSuccessListener
         $singleUse,
         RefreshTokenGeneratorInterface $refreshTokenGenerator,
         ExtractorInterface $extractor,
-        array $cookieSettings
+        array $cookieSettings,
+        $returnExpiration,
+        $returnExpirationParameterName
     ) {
         $this->refreshTokenManager = $refreshTokenManager;
         $this->ttl = $ttl;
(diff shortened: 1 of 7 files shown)


EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -59,10 +59,22 @@ class AttachRefreshTokenOnSuccessListener
 
     protected array $cookieSettings;
 
+    /**
+     * @var bool
+     */
+    protected $returnExpiration;
```
mbabker, line 65:
```suggestion
    protected bool $returnExpiration;
```

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -59,10 +59,22 @@ class AttachRefreshTokenOnSuccessListener
...
+
+    /**
+     * @var string
+     */
+    protected $returnExpirationParameterName;
```
mbabker, line 70:
```suggestion
    protected string $returnExpirationParameterName;
```

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -72,7 +84,9 @@ public function __construct(
         $singleUse,
         RefreshTokenGeneratorInterface $refreshTokenGenerator,
         ExtractorInterface $extractor,
-        array $cookieSettings
+        array $cookieSettings,
+        $returnExpiration,
+        $returnExpirationParameterName
```
mbabker, line 89: Adding new required constructor params is a B/C break, they must be optional.

```suggestion
        bool $returnExpiration = false,
        string $returnExpirationParameterName = 'refresh_token_expiration'
```

README.md
```diff
@@ -220,6 +220,22 @@ gesdinet_jwt_refresh_token:
     token_parameter_name: refreshToken
 ```
 
+### Return Expiration Timestamp
+
+If set to true, the expiration timestamp will be added to the response.
```
mbabker, line 225: Perhaps also note that the timestamp is a Unix timestamp so folks don't expect a formatted DateTime.

mbabker (changes requested):

EventListener/AttachRefreshTokenOnSuccessListener.php
ricardodevries: √

EventListener/AttachRefreshTokenOnSuccessListener.php
ricardodevries: √

EventListener/AttachRefreshTokenOnSuccessListener.php
ricardodevries: √

README.md
ricardodevries: √

mbabker (approved):

Issue #316, opened 2022-04-13
Question: refreshing invalid JWT with valid refresh token?


sebastianstucke87:
Hello,

According to your documentation, the JWT must still be valid in order to be renewed (with a valid refresh token ofc):
> Ask to renew **valid JWT** with our refresh token. Make a POST call to (...)
>
> (via https://github.com/markitosgv/JWTRefreshTokenBundle#generating-tokens)

Is that correct? Because if so, oh boy! That would be quite a bummer... 😅

mbabker:
Nah, that's just bad wording.  As long as the refresh token is valid then you're good; on a successful refresh, the refresh endpoint generates a new JWT.

geoffroyp:
Are you sure about that? Because I'm struggling at the moment with this: If I call the refresh route with a valid-but-expired JWT token, and a valid and not expired refresh token, I get a 401 ExpiredJWT token because in this case, `gesdinet.refresh_token` is never called.

With the exact same API code, if I now call the refresh route with a valid and not expired JWT token, `gesdinet.refresh_token` is called, and everything works fine...

mbabker:
An expired JWT can’t be considered valid in the same way an expired session can’t be resumed.

Very loosely, the refresh token can be compared to a remember me cookie; it’s only issued after the user authenticates their self to the application, and it is only valid for that user based on how you configure the remember me feature.

webda2l:
Related https://github.com/markitosgv/JWTRefreshTokenBundle/issues/334

AimSai59:
I have indeed doubts about the security of this system... I can create new JWT only with the refresh token. Let's imagine that my database leaks, anyone can connect with the account of each of my users, just by using the refresh token found in the database. 

He just has to go to /api/token/refresh, give the refresh token, and he has a valid JWT ! I don't understand the use of hashing the passwords if it's to display the refresh tokens in clear.... Maybe a hash of the tokens in database would be a good thing. Only the user who received the refresh token in clear text during his connection can use it...

mbabker:
> I have indeed doubts about the security of this system...

If your database leaks, the refresh token values being stored in plaintext is the least of your concerns.

While #289 would be a better spot to chime in with your concerns, let me explain the problem here.

[...]

PR #318, opened 2022-04-15
Update documentation (#316)

sebastianstucke87:
please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -545,12 +545,12 @@ When you authenticate through /api/login_check with user/password credentials, L
 }
 ```
 
-The refresh token is persisted as a `RefreshTokenInterface` object. After that, when your JWT valid token expires, if you want to get a new one you can proceed in two ways:
+The refresh token is persisted as a `RefreshTokenInterface` object. When your JWT expires, you have two options:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -545,12 +545,12 @@ When you authenticate through /api/login_check with user/password credentials, L
 }
 ```
 
-The refresh token is persisted as a `RefreshTokenInterface` object. After that, when your JWT valid token expires, if you want to get a new one you can proceed in two ways:
+The refresh token is persisted as a `RefreshTokenInterface` object. When your JWT expires, you have two options:
 
-- Send you user credentials again to /api/login_check. This generates another JWT with another Refresh Token.
-- Ask to renew valid JWT with our refresh token. Make a POST call to /api/token/refresh url with refresh token as payload. In this way, you can always get a valid JWT without asking for user credentials. But **you must check** if the refresh token is still valid. Your refresh token will not change but its TTL will increase.
+- Generate a new JWT by re-authenticate with your credentials via `/api/login_check`. This will also generate a new refresh token.
```
webignition, line 550: "... by re-authenticating with ..." reads more correctly.

shakaran:
@sebastianstucke87 Thanks for your contribution!

PR #325, opened 2022-05-31
Fix token expiration date

ricardodevries:
Hello maintainers,

Instead of using `time() + $this->ttl` my suggestion is to use the `getValid()` method of the token. If you request a new token through `/api/token/refresh`, you get the latest unix time while the token may expire another timestamp. This fix uses the getValid method and turns it into an timestamp.

As always I am open to suggestions, critical feedback leading to improvements on this PR or comments 👍


ricardodevries:
I'm checking up to see if anything else is needed to finish and merge this PR.

ricardodevries:
Just checking in.

I was wondering if you’d had a chance to review my PR?

Thanks!

ricardodevries:
Just following up. I apologize in advance for bothering you.

PR #328, opened 2022-06-15
Fix deprecations on Symfony 6.1

gordinskiy:
Fix deprecation notice:

> Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute instead.

### DIFF
--- a/Command/ClearInvalidRefreshTokensCommand.php
+++ b/Command/ClearInvalidRefreshTokensCommand.php
@@ -13,14 +13,19 @@ namespace Gesdinet\JWTRefreshTokenBundle\Command;
 
 use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenInterface;
 use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManagerInterface;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
+#[AsCommand(name: 'gesdinet:jwt:clear', description: 'Clear invalid refresh tokens.')]
 class ClearInvalidRefreshTokensCommand extends Command
 {
+    /**
+     * @deprecated
+     */
     protected static $defaultName = 'gesdinet:jwt:clear';
 
     private RefreshTokenManagerInterface $refreshTokenManager;
--- a/Command/RevokeRefreshTokenCommand.php
+++ b/Command/RevokeRefreshTokenCommand.php
@@ -12,14 +12,19 @@
 namespace Gesdinet\JWTRefreshTokenBundle\Command;
 
 use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManagerInterface;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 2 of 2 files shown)


maxhelias (approved):

Command/ClearInvalidRefreshTokensCommand.php
```diff
@@ -13,14 +13,19 @@
...
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
+#[AsCommand(name: 'gesdinet:jwt:clear', description: 'Clear invalid refresh tokens.')]
 class ClearInvalidRefreshTokensCommand extends Command
 {
+    /**
+     * @deprecated
```
ricardodevries, line 27: Attributes are PHP 8+. Does this mean the package will drop support for PHP 7.4?

maxhelias: No need, attributes do not break on older versions, they are not taken into account

KDederichs: that'll bite you on PHP 7.4 since it can't parse the use Symfony\Component\Console\Attribute\AsCommand due to syntax

KDederichs:
Best way (I think) to remove that deprecation while still maintaining BC is overwriting the `getDefaultName()` function. That way it works in all PHP and SF versions.

Command/ClearInvalidRefreshTokensCommand.php
KDederichs: The problem is that 5.4 does have the Attribute:

https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Console/Attribute/AsCommand.php

And that does support 7.4 :)

DmytryHo:
I tried to make changes consistent with changes from [lexik/LexikJWTAuthenticationBundle](https://github.com/lexik/LexikJWTAuthenticationBundle)
`$defaultName` already marks as deprecated there: [Fix deprecations on Symfony 6.1 ](https://github.com/lexik/LexikJWTAuthenticationBundle/commit/a44912a7ec882f9b85d27376258228c1a8caad86)

KDederichs:
Very strange, for some reason it blew up when I tried fixing this deprecation for api-platform and then including the `AsCommand`

Command/ClearInvalidRefreshTokensCommand.php
mbabker: The attribute syntax was explicitly chosen because it is parsed as a comment on PHP 7 and earlier.  Adding attributes to a class will not break PHP 7 compatibility.  The only thing this approach does is cause duplication of information in order to keep the command class lazy-loaded.

KDederichs: Yeah, I just had a weird thing happening where I tried to import `AsCommand` as well on Api-Platform to fix the deprecation there and it basically blew up in my face complaining about the file (assuming cause of the constructor syntax) so I'm just a bit surprised it doesn't seem to do it here/ on the Lexik one.
Maybe I'll try to dig into that some more, just strikes me as odd.

imdhemy:
@markitosgv 
Could you please merge this PR? I understand you may be busy, but I would really appreciate it if you could find the time to do so. Thanks! 🙌

chalasr (approved):
LGTM as-is

mortezakarimi (approved):

thomaspicquet:
time to get this merged, pretty please ❤️

mortezakarimi (approved):

bpolaszek:
🆙 🙏

FabienVINCENT:
pray for merge :pray:

Orkin:
Any news here ... ?

PR #344, opened 2022-11-10
Fix php 7.4 pipelines

eerison:
https://github.com/markitosgv/JWTRefreshTokenBundle/actions/runs/3355047594/jobs/5559029249

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
   },
   "require-dev": {
     "doctrine/cache": "^1.11|^2.0",
-    "doctrine/mongodb-odm": "^2.0",
+    "doctrine/mongodb-odm": "^2.2",
     "doctrine/orm": "^2.7",
     "matthiasnoback/symfony-config-test": "^4.2",
     "matthiasnoback/symfony-dependency-injection-test": "^4.2",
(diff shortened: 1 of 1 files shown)


maxhelias:
I think we should bump the version of doctrine/mongodb-odm to `^2.2` instand of `^2.0` which no longer has the package

eerison:
even if I bump to `"doctrine/mongodb-odm": "^2.2",`, composer ask me to add `"ocramius/package-versions": true`

eerison:
you are correct! I missed to add `--prefer-stable --prefer-lowest --prefer-dist`

maxhelias (approved):

mbabker:
You should update the conflict as well otherwise the minimum version being tested in dev doesn't match up with the minimum version that can be installed in production.

eerison:
> You should update the conflict as well otherwise the minimum version being tested in dev doesn't match up with the minimum version that can be installed in production.

Ahhh I got it know :D 

because it's only in require-dev, yeah it's needed for conflicts as well!

mbabker (approved):

shakaran:
@mbabker could be this merged to master? Thanks

thomaspicquet:
@markitosgv let us know if you need extra help with this bundle so PRs can get merged faster

Issue #343, opened 2022-10-20
[RFC] 2.0 Proposal


mbabker:
I've been sitting on a branch for a while that has a bunch of proposed changes for a 2.0 branch.  You can see the diff [here](https://github.com/markitosgv/JWTRefreshTokenBundle/compare/master...mbabker:JWTRefreshTokenBundle:2.0-proposal) and the relevant [upgrade guide](https://github.com/mbabker/JWTRefreshTokenBundle/blob/2.0-proposal/UPGRADE-2.0.md).

To sum it up though:
- Minimum supported Symfony version is now 5.4
- Minimum supported PHP version is now 8.1 (the [Packagist stats for the 1.x releases](https://packagist.org/packages/gesdinet/jwt-refresh-token-bundle/php-stats#1) show this covers ~70% of users)
- Deprecated functionality dropped
- Typehints everywhere

So, it's not a huge release as far as major versions go.  But, given Symfony 4.4 is about to stop receiving regular maintenance and go into LTS security-only support, this is a good time to look at what dropping support for it looks like.  As for the PHP minimum, 7.4 reaches its end of support in a couple of months and Symfony 6 uses 8.1 as a minimum, so these versions would match up nicely.  And for those still on older Symfony or PHP versions, the 1.x branch is still there and totally usable.

geoffroyp:
Everything makes sense in my opinion

Go for it! :)

thomaspicquet:
@mbabker great work, would love to see this new 2.0 version started, let me know if you want a beta tester

ger86:
Great @mbabker . Can I help you in accomplishing this update?

mbabker:
PR @ #347

PR #347, opened 2022-12-01
2.0 Proposal

mbabker:
See #343 for details

I'm leaving it as split-up commits to make it a little easier to see how everything evolved.

If accepted as is, what I'd personally do is this:

- Merge these PRs:
    - [x] #328
    - [x] #344
    - [x] #348
    - [x] #350
- Cut a 1.4 release
- Merge this
- Cut a 2.0 beta


DependencyInjection/Configuration.php
```diff
@@ -108,10 +71,6 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->scalarNode('remove_token_from_body')->defaultTrue()->end()
                     ->end()
                 ->end()
-                ->scalarNode('logout_firewall')
```
maxhelias, line 111: We should depreciate it before we delete it

EventListener/LogoutEventListener.php
```diff
@@ -43,55 +42,51 @@ public function __construct(
...
+
         if (null === $tokenString) {
             $event->setResponse(
                 new JsonResponse(
                     [
                         'code' => 400,
                         'message' => 'No refresh_token found.',
                     ],
-                    JsonResponse::HTTP_BAD_REQUEST
+                    Response::HTTP_BAD_REQUEST
```
maxhelias, line 60: BTW, we should not overload a new response if one exists and let the "DefaultLogoutListener" set it
And also set a negative priority to be before the "CookieClearingLogoutListener"

DependencyInjection/Configuration.php
mbabker: Yeah, thanks for the reminder here.  There are definitely a few things that could use a proper deprecation in a 1.x release that I'm now forgetting about since I've been sitting on this branch for so long 😬

EventListener/LogoutEventListener.php
mbabker: > BTW, we should not overload a new response if one exists and let the "DefaultLogoutListener" set it

Well, this is a bit of a chicken vs. egg problem.  The `DefaultLogoutListener::onLogout()` listener has a priority of 64, which makes it the first event listener to run on the logout event in the app for one of my company's bigger clients.  So an `if ($event->getResponse() === null)` check isn't going to have a lot of value here without making it a higher priority than the default listener.

> And also set a negative priority to be before the "CookieClearingLogoutListener"

The `CookieClearingLogoutListener::onLogout()` listener already has a priority of -255, so this listener is already running before that one (when it's enabled).

maxhelias: Tell me if I'm wrong: in case your application uses a form_login on the same firewall as the refresh_jwt, the `{firewall}.logout.path` configuration will not work anymore because the response created by the `DefaultLogoutListener` will be overwritten by this listener.
And BTW, even in the current configuration of this bundle, we should have this case

mbabker: In that scenario, yeah, you're hosed.

This almost needs a `generate_response` config param TBH, one can then explicitly opt into whether this bundle generates (and potentially overwrites a previously set) response for the logout request.

You then need to split the cookie invalidation part of the listener off into another lower priority listener (maybe the same priority as `CookieClearingLogoutListener::onLogout()`) that way it runs as late in the process as possible and has a better chance of clearing the cookie on the right response object (in case another bundle or a listener in the app itself is doing something with the response).

maxhelias: Yeah a config on the refresh_jwt like `logout_path` to check if the route match to generate the json response or not

thomaspicquet (approved):

DependencyInjection/Configuration.php
mbabker: #350 for compat layer

Request/Extractor/RequestBodyExtractor.php
```diff
@@ -17,7 +17,7 @@ final class RequestBodyExtractor implements ExtractorInterface
 {
     public function getRefreshToken(Request $request, string $parameter): ?string
     {
-        if (null === $request->getContentType() || false === strpos($request->getContentType(), 'json')) {
+        if (null === $request->getContentType() || !str_contains($request->getContentType(), 'json')) {
```
thomaspicquet, line 20:
```suggestion
        $format = method_exists(Request::class, 'getContentTypeFormat') ? $request->getContentTypeFormat() : $request->getContentType();
        if (null === $format || !str_contains($format, 'json')) {
```

shakaran:
@thomaspicquet Could you rebase the conflicts to keep this big one issue updated? Thanks

thomaspicquet:
@mbabker I've been using your version 2.0 on production for a couple months now and its been working well ;)
if that helps to get this merged :)

shakaran:
Please @markitosgv @mbabker, could we merge this for get rid of a lot deprecation warnings annoying? The 2.0 release is taking a lot time to happen.

We need remove the deprecations and old code and release the 2.0 clean for Symfony 7.0 world. It is very annoying see the deprecations warnings from 1.0 release or master.

All the uses from 

`use Gesdinet\JWTRefreshTokenBundle\Generator\RefreshTokenGeneratorInterface;`

Instead 

`use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManagerInterface;`

Are spamming a lot project logs with deprecated.

Thanks

mbabker:
> We need remove the deprecations and old code and release the 2.0 clean for Symfony 7.0 world. It is very annoying see the deprecations warnings from 1.0 release or master.

What deprecations, specifically, are you hitting with this bundle's latest 1.x release that cannot be fixed?  The bundle itself shouldn't be using its own deprecated code.  If your issue is just the presence of `@deprecated` annotations, that alone isn't a problem (the Symfony debug loader won't randomly spit out log messages for deprecated methods when they aren't called).

> All the uses from
> 
> `use Gesdinet\JWTRefreshTokenBundle\Generator\RefreshTokenGeneratorInterface;`
> 
> Instead
> 
> `use Gesdinet\JWTRefreshTokenBundle\Model\RefreshTokenManagerInterface;`
> 
> Are spamming a lot project logs with deprecated.

Is something in your project using the deprecated `RefreshTokenManagerInterface::create()` method?  Outside of the test suite, nothing in the bundle is calling it.

mbabker:
> it some way, when I load the bundle, always hit https://github.com/markitosgv/JWTRefreshTokenBundle/blob/master/Entity/AbstractRefreshToken.php#L16 since the autoloaders with PSR4 load all the claseses.

It's no different than `trigger_deprecation()` calls that live outside classes in Symfony files; files like the deprecated `ContainerAwareTrait` or `RequestMatcher` class would cause the same type of excess deprecation logging if they were arbitrarily included as well.

That deprecation being triggered isn't the autoloader arbitrarily including all files, but rather processes that eagerly scan and load files.  I know API Platform has stuff that can do that (exactly as pointed out by that issue), moving those calls doesn't necessarily solve anything here (and if anything it's a higher risk change to introduce a constructor to an abstract class where one didn't exist previously).

shakaran:
I tag and release 1.5.0 with previous changes. After I merge #350. So if this PR get rebased or still missing to merge something? The idea is create a new 2.0.0-dev prerelease and tag. All PR are welcome

mbabker:
This is as cleaned up as I can get it for now.  I've un-drafted it, but probably could use some good testing and review before a merge.

PR #348, opened 2022-12-06
[1.x] Adjust CI setup to get functional MongoDB install again

mbabker:
Extracted from #347

With GitHub updating the `ubuntu-latest` alias to 22.04, the `mongodb-labs/drivers-evergreen-tools` action can't set up a MongoDB server correctly anymore.  It doesn't look to support the latest Ubuntu LTS, so this downgrades the Ubuntu version.

The MongoDB server version is upgraded as well.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


mbabker:
Prefer lowest job is fixed by #344

PR #349, opened 2022-12-06
[1.x] Fix PHP 8.2 compat issue in uses of `DateTime::modify()`

mbabker:
As noted in https://github.com/php/php-src/issues/9950 there is a behavioral change in PHP 8.2 if you make a call like `$dateTime->modify('+-300 seconds')` as a result of another bug fix in the date/time handling libraries.  As the bundle does not enforce TTLs to be a positive number (and the functional tests abuse this a bit), this is a practical fix to ensure anything passing a negative TTL value doesn't catastrophically break.

### DIFF
--- a/Model/AbstractRefreshToken.php
+++ b/Model/AbstractRefreshToken.php
@@ -41,7 +41,13 @@ abstract class AbstractRefreshToken implements RefreshTokenInterface
     public static function createForUserWithTtl(string $refreshToken, UserInterface $user, int $ttl): RefreshTokenInterface
     {
         $valid = new \DateTime();
-        $valid->modify('+'.$ttl.' seconds');
+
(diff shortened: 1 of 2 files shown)


mbabker:
Functional CI blocked by #348

maxhelias (approved):

shakaran:
@maxhelias any update missing for merge this to master? Compat with 8.2 is very important for next upgrades. Thanks

maxhelias:
@shakaran I am not the owner or a collaborator. Thanks

mbabker:
I don't have merge rights here.

shakaran:
> I don't have merge rights here.

who has? By contributions maybe @markitosgv (composer author) or @paulchubatyy ?

thomaspicquet (approved):

mbabker:
Superseded by #361

PR #368, opened 2023-11-14
feat: add support for partitioned cookies

### LINKED ISSUE #367: Support for partitioned authentication cookies
Hey folks!

With the upcoming [phasing-out of third-party cookies by Chromium](https://developer.chrome.com/docs/privacy-sandbox/third-party-cookie-phase-out/) (and, I expect, other browser vendors in the near future), I think it would be useful to add support for Partitioned authentication cookies in this library.

Some applications embed widgets (or even their entire application!) on other sites in a legitimate way. In these cases, their authentication cookies are usually set with `SameSite=None`, and considered as third-party cookies by browsers. The upcoming phase-out will therefore break any such application... unless they switch to partitioned cookies (see Google's docs on [CHIPS](https://developer.chrome.com/docs/privacy-sandbox/chips/) - _Cookies Having Independent Partitioned State_).

[Symfony has added support for Partitioned Cookies in version 6.4](https://symfony.com/blog/new-in-symfony-6-4-chips-cookies) (which will be released by the end of the month). This means that this library could fairly easily be updated to also allow this option, by simply adding an additional configuration option, which defaults to the current behavior (not partitioned).

Here is what an example configuration for an application in this context could look like, once this is implemented:

```yaml
gesdinet_jwt_refresh_token:
    # ...

    cookie:
        enabled: true
        same_site: none
        path: /auth/refresh
        domain: null
        http_only: true
        secure: true
        partitioned: true       # <--- new option
        remove_token_from_body: true
```

If the maintainers do not have any objections, I could get to work on a PR for this, as we do have a need for this in a project that I am currently working on (in order to give us a bit of peace of mind regarding the upcoming phase out 😅). 

Let me know if you have any questions, objections or feedback! Cheers!

EmilePerron:
This PR adds support for partitioned third-party cookies in a backwards compatible way, as described in #367.

Let me know if you have any comments/feedback!

closes #367

### DIFF
--- a/EventListener/AttachRefreshTokenOnSuccessListener.php
+++ b/EventListener/AttachRefreshTokenOnSuccessListener.php
@@ -19,6 +19,7 @@ use Lexik\Bundle\JWTAuthenticationBundle\Event\AuthenticationSuccessEvent;
 use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpKernel\Kernel;
 
 class AttachRefreshTokenOnSuccessListener
 {
@@ -99,6 +100,10 @@ class AttachRefreshTokenOnSuccessListener
         ], $cookieSettings);
         $this->returnExpiration = $returnExpiration;
         $this->returnExpirationParameterName = $returnExpirationParameterName;
+
+        if ($this->cookieSettings['partitioned'] && Kernel::VERSION < '6.4') {
+            throw new \LogicException(sprintf('The `partitioned` option for cookies is only available for Symfony 6.4 and above. You are currently on version %s', Kernel::VERSION));
+        }
     }
 
     public function attachRefreshToken(AuthenticationSuccessEvent $event): void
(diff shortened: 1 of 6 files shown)


EmilePerron:
Hello! Would it be possible to know when we can expect this to be reviewed and/or merged, or if there is anything prevent this from moving forward? :) 

(Got a thumbs up from a fellow Github user, so we know at least one person other than my colleagues and I seems to be interested in this 😅)

maxhelias (approved):

maxhelias:
Interesting feature indeed!

EventListener/AttachRefreshTokenOnSuccessListener.php
```diff
@@ -160,7 +166,8 @@ public function attachRefreshToken(AuthenticationSuccessEvent $event): void
                     $this->cookieSettings['secure'],
                     $this->cookieSettings['http_only'],
                     false,
-                    $this->cookieSettings['same_site']
+                    $this->cookieSettings['same_site'],
+                    $this->cookieSettings['partitioned']
```
webignition, line 170: If you add a trailing comma at the end of this line,  future changes that add a line will only need to edit one line and not two.

Hardly a necessary change but perhaps something worth considering.

EmilePerron: Fair point! 

I hadn't done this because I wasn't sure which PHP version added support for trailing commas and didn't want it to break on older versions, but I just looked it up and it was added in 7.3. 

Seeing as this package supports `>= 7.4`, I just added the trailing comma :)

EmilePerron:
Hi @markitosgv! 

Just got back to work after the holidays - _hope you and everyone else here had a good time by the way_ - and was wondering if there is anything I could do to get this PR moving? :)

Issue #369, opened 2023-12-04
Fatal error on latest version with Symfony 7


gordinskiy:
Updating to Symfony 7+ leads to fatal error:

> Fatal error: Declaration of Gesdinet\JWTRefreshTokenBundle\Security\Exception\InvalidTokenException::getMessageKey() must be compatible with Symfony\Component\Security\Core\Exception\AuthenticationException::getMessageKey(): string in /opt/project/vendor/gesdinet/jwt-refresh-token-bundle/Security/Exception/InvalidTokenException.php on line 21

**gesdinet/jwt-refresh-token-bundle**: v1.2.0
**Symfony/***: v7.0.1

mbabker:
https://github.com/markitosgv/JWTRefreshTokenBundle/pull/366

PR #366, opened 2023-11-13
Fix SF 7 compat

mbabker:
The [Symfony 7 compat PR](https://github.com/markitosgv/JWTRefreshTokenBundle/pull/365) was merged and released too quickly, given that:

- ~`lexik/jwt-authentication-bundle` itself [has a pending PR](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1165) for Symfony 7 support~ Merged
- The original PR added  `symfony/security-guard` versions that do not exist to the Composer manifest
- The original PR didn't add CI coverage for Symfony 7 (or 6.4), so the passing CI is a false positive since that combination is untested

~This PR is a draft PR until the `lexik/jwt-authentication-bundle` PR is merged, as that will block the Symfony 7 build from running until it's merged/released.~ Issues fixed and ready for review/merge.

PR #370, opened 2023-12-05
fix: return type issue for php8.1

saysource:
Without setting the return type in the exceptions, I get the error already logged here:

#369 

Hope this helps, it is a really simple fix.


webignition:
This is a subset of the changes in PR #366. 

Perhaps this should be closed in preference to merging #366 as that also includes additional SF7 compatibility changes.

saysource:
Yep, closing.

PR #381, opened 2024-05-13
Drop support for Symfony 4.4

mbabker:
Symfony 4.4 is EOL, so just as the title says, this drops support for the legacy branch.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -10,13 +10,13 @@ jobs:
       fail-fast: false
       matrix:
         php: ['7.4', '8.0', '8.1', '8.2', '8.3']
-        symfony: ['4.4.*', '5.4.*', '6.0.*', '6.1.*', '6.2.*', '6.3.*', '6.4.*', '7.0.*']
+        symfony: ['5.4.*', '6.0.*', '6.1.*', '6.2.*', '6.3.*', '6.4.*', '7.0.*']
         composer-flags: ['--prefer-stable']
         can-fail: [false]
         extensions: ['curl, iconv, mbstring, mongodb, pdo, pdo_sqlite, sqlite, zip']
         include:
           - php: '7.4'
-            symfony: '4.4.*'
+            symfony: '5.4.*'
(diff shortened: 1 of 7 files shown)


okorneliuk:
@mbabker please add 7.1 to the matrix and mention it in the title of the PR.

Also, CI does not like this:

`use Symfony\Component\Security\Http\RememberMe\RememberMeHandlerInterface;`

mbabker:
> please, drop 6.0 - 6.3 as well, because these versions are also EOL. Please, add 7.1 to the matrix.

Neither of these changes is related to this pull request.  Also, from a practical perspective, dropping the 6.0-6.3 releases doesn't change anything meaningful in the code given that 5.4 is still supported (the only thing it does is get rid of a few CI builds).  On the other hand, dropping 4.4 support allows removing some B/C-related code.

> Also, CI doesn't like this:

The CS issue is fixed.

okorneliuk:
@mbabker You are right, these changes are not related to the current title of the PR. However, it would be convenient to reorder versions once, because after this change, major version of the bundle should be bumped according to SemVer. Another dropping will cause another bump, but as you mentioned, dropping of 6.0-6.3 is not so significant. You can edit the title of the PR to reflect all changes you've made.
If you are still against doing these changes here, I will likely create a new PR. Thank you for your contribution :+1:

mbabker:
> However, it would be convenient to reorder versions once, because after this change, major version of the bundle should be bumped according to SemVer.

Dropping support for Symfony 4.4 does not require a SemVer major release, the API of this bundle has not changed, and a package manager will properly restrict updates on unsupported environments.  As a convenience, many package maintainers will not drop support for environments without a major version release, but this is not a hard requirement.

The pull request title and description are accurate for the focus of this pull request.  Further pull requests can target other changes, but lumping unrelated changes together in a single pull request just makes review more complicated (i.e. how does adding a test build for Symfony 7.1, and addressing potential deprecations that are found from that, relate to dropping support for Symfony 4.4?).

maxhelias (approved):

PR #401, opened 2025-06-26
Use modern filesystem layout

mbabker:
Updates the filesystem layout to use modern bundle suggestions, including:

- Config moved to top level instead of Resources
- Source code in `src`


shakaran (approved):

shakaran:
@mbabker thanks for your contribution!

PR #413, opened 2025-11-12
feat: Added support for plain dbal

jorgerobles:
Features a plain DBAL adapter
* Optimized queries beyond object manager
* Custom table and columns via configuration.


jorgerobles:
@shakaran Is necessary to support largely deprecated and unmaintained packages for the new 2.0 version?

Doctrine introduced several BC, in Connection and QueryBuilder

Doctrine < 3.6
`$connection->createSchemaManager())`

Doctrine 2.x
`$querybuilder->execute()->fetch(PDO::FETCH_ASSOC)`

shakaran:
Thanks @jorgerobles, and sorry for the long wait. The feature is worth having: a bundle that needs neither the ORM nor the ODM to store its tokens is a reasonable thing to offer, and the implementation is careful. Rather than send back another list, I brought the branch up to date and pushed the changes on top of yours. All 19 checks pass now and it is mergeable.

Worth saying first: your batch revocation is right where the Doctrine one was wrong. It keeps the offset where it is between batches, with a comment saying why, and it accumulates what it revoked. The Doctrine manager did neither until earlier today, which meant it skipped a batch for every batch deleted and always reported that it had revoked nothing. You got that right in November.

**What I changed, and why**

*Table creation.* It was on by default and ran on every request until it succeeded, and the `catch (\Throwable)` logged the failure and carried on. That means the connection serving traffic needs rights to alter the schema, which is not something to opt someone into, and a permissions problem would surface later as an unrelated missing table. It is off by default now, documented as a fallback for when a migration is not practical, and a failure stops the request with a message saying what to do.

*Hydration.* The bound closure wrote straight into the private properties of `AbstractRefreshToken`. That ties the manager to those property names, and although the constructor accepts `class-string<RefreshTokenInterface>`, it only worked for classes extending that one. It goes through the setters now, with reflection left for the identifier alone, since that is the only field without one, and skipped when a model has no such property.

*Identifier quoting.* Reads quoted the table and columns, writes did not. A column named `order`, or anything else needing quotes, would read fine and fail on write.

*`save()`.* It read the row back before every write to choose between insert and update. Updating first answers the same question, and it is one query instead of two whenever the token is already there. Your identifier validation I kept, though I rewrote it without `preg_match`, which phpstan treats as impure while psalm wanted the method marked pure.

*Two things I ran into.* The branch had dropped the `RuntimeException` for a missing object manager, so a misconfigured application failed later on an unresolvable service instead. And the mutual exclusion check read `$v['object_manager']` without checking the key was there, which warned on a partial configuration.

*Quality gates and docs.* phpstan runs at level 10 and psalm is clean on the branch. The suite is at 185 tests, with new ones covering the identifier surviving a round trip, saving the same token twice updating in place, and the table creation failure being reported. The backend was not documented anywhere, so the readme has a section on it, including what turning the auto creation on actually means, and the changelog mentions it.

**Left as you wrote it**

The DBAL manager stays a separate implementation. Sharing code with the Doctrine one would mean an `ObjectManager` adapter over a connection with no unit of work, which is a lot of surface to fake for little gain. What I would rather share is the tests: one contract suite run against all three backends would have caught the batch bugs on the Doctrine side, given yours was already correct. That is a separate piece of work.

Merging this. Thanks for the patience, and for the care in the implementation.

PR #424, opened 2026-03-17
Fix token revocation listing for tokens when using ODM

melkamar:
When using ODM, the result is not an array but
a CachingIterator. That needs to be converted
to an array for the array_map method, else it
crashes:

```
In ClearInvalidRefreshTokensCommand.php line 67:

  array_map(): Argument #2 ($array) must be of type array, Doctrine\ODM\MongoDB\Iterator\CachingIterator given

```

Also, if this could be patched as 1.5.x, that would be great as we are not in 2.x 🙏

### DIFF
--- a/src/Command/ClearInvalidRefreshTokensCommand.php
+++ b/src/Command/ClearInvalidRefreshTokensCommand.php
@@ -66,7 +66,7 @@ final class ClearInvalidRefreshTokensCommand extends Command
             $io->text(sprintf('Revoked %d invalid token(s) in batches of %d', count($revokedTokens), $batchSize));
             $io->listing(array_map(
                 static fn (RefreshTokenInterface $revokedToken): string => $revokedToken->getRefreshToken(),
-                $revokedTokens,
+                iterator_to_array($revokedTokens),
             ));
         }
 


shakaran:
@melkamar could you create a test that reproduce the previous case and new case? So I can backport easily to 1.5.x. Thanks

shakaran:
Thanks @melkamar, and sorry for the delay. I looked into this in depth: the bug is real, but the fix belongs one layer lower, so I handled it a bit differently.

**On 1.5.x (your case)**

You're right. There, `RefreshTokenManager::revokeAllInvalid()` has no return type, so the ODM's `CachingIterator` is passed straight through to the command and `array_map()` blows up exactly as you reported.

One caveat about the patch itself: 1.5.x supports `php >=7.4`, and `iterator_to_array()` only started accepting arrays in PHP 8.2. On 7.4–8.1 it would turn the crash around and break the ORM path instead, where the repository does return a plain array. The normalization needs an `is_array()` guard.

So I fixed it in the manager rather than in the command, which means every consumer gets the `RefreshTokenInterface[]` that `RefreshTokenManagerInterface` documents, not just this one command: bca2e7b, on the new `1.5.x` branch. Released as **v1.5.1**.

About the tests I asked you for: I wrote them along with the fix. The functional one wires the real manager to a repository returning an iterator, and before the fix it reproduces your traceback verbatim, same file and same line (`ClearInvalidRefreshTokensCommand.php line 67`).

**On master (2.x)**

The change doesn't apply here. `RefreshTokenManagerInterface::revokeAllInvalidBatch()` is typed `: array`, so a `CachingIterator` can never reach that `array_map()` — the manager would throw a `TypeError` on its `return` first. That is also why the test can't be written at the command level on 2.x: PHPUnit mocks honour the return type.

Digging into that path did surface three worse bugs on 2.x, now fixed in 83284d6 and f153a20:

* `revokeAllInvalidBatch()` returned the last batch, which is always empty when the loop exits, so the command reported "There were no invalid tokens to revoke" even after deleting thousands of them.
* The loop condition was `while (!empty($invalidTokens))`, and `empty()` is never true for an object, so with the ODM it never terminated.
* The offset moved forward on every iteration even though each batch is deleted before the next one is read, so one batch was skipped for every batch deleted.

I'm closing this PR in favour of those commits, but the credit for spotting it is yours. Thanks for the report!

PR #426, opened 2026-04-24
Fix README badges

manuelmeister:


shakaran (approved):

shakaran:
@manuelmeister even if removes the scrutinizer, which I dont have the admin access, probably it is better that keep the 404 image ;) Thanks

PR #429, opened 2026-06-07
feat: add token revocation by username

Rainrider:
This PR adds refresh token revocation by username in a backwards compatible manner.

## Rationale

I would like to be able to revoke all refresh tokens issued to a specific user upon password reset in a performant way. This is currently not possible using the `RefreshTokenManager` but could be implemented client-side using something like that:

```php
/** @var \Doctrine\ORM\EntityManagerInterface $this->manager */
/** @var \Symfony\Component\Security\Core\User\UserInterface $user */

$repo = $this->manager->getRepository(RefreshToken::class);
$repo->createQueryBuilder('rt')
    ->delete(RefreshToken::class, 'rt')
    ->where('rt.username = :user_identifier')
    ->setParameter('user_identifier', $user->getUserIdentifier())
    ->getQuery()
    ->execute();
```

This is however coupled to the underlying database type (SQL vs NoSQL). As the `RefreshTokenManager` abstracts away the database type, this would be better implemented there.

## Implementation details

- Two new interfaces (`DeleteRefreshTokenRepositoryInterface` and `RevokeRefreshTokenManagerInterface`) have been added to address backwards compatibility
- The refresh token deletion is implemented using DQL and is performed directly by the database, thus no refresh token objects are added to the unit of work and no life-cycle events are raised. This should be the most performant deletion strategy as no object hydration is performed by doctrine (other than the query result).

## Extension considerations

Expired refresh tokens are currently deleted in batches to address memory issues. However the deleted refresh token objects are still hydrated by doctrine and life-cycle events are raised for them. A more performant method could be used like the one from this PR when people do not care about which tokens were exaclty revoked and do not rely on life-cycle events being raised upon token deletion. Adding such a method after release would however necessitate either more interfaces or break backwards compatibility. Please tell me if you like this, I might then add something like `RevokeRefreshTokenManagerInterface::revokeAllExpired` before you release a new version of the library.


Rainrider:
Please see #430 for the failed Psalm Security Workflow. The failure is not related to this PR.

shakaran:
Thanks @Rainrider, and sorry this sat for so long. The feature makes sense and the design is right: revoking every token of a user is what a password reset needs, and keeping it behind separate interfaces leaves existing implementations alone.

Master moved a lot in the meantime, so rather than send you back a list, I rebased your branch and pushed two commits on top of yours. Your commit is unchanged in content; the only conflict was the test file, where we had both appended a method. All 19 checks pass now and the branch is mergeable.

**What I pushed**

*Quality gates adopted while this was open*, so the analysis runs clean:

* `#[\Override]` on the three new methods, and the purity annotations psalm now asks for on interfaces
* `deleteByUser()` in the ORM repository returned `mixed` at phpstan level 10, since `Query::execute()` is untyped
* The docblock on `RevokeRefreshTokenManagerInterface` was the one from `RefreshTokenManagerInterface` and described something else, so it now says what the interface is for and why it is separate

*Tests for the manager.* Only the repositories were covered, so `revokeAllForUser()`, which is the method an application actually calls, was not exercised at all, and neither was the `LogicException` a repository predating `DeleteRefreshTokenRepositoryInterface` runs into. Both are covered now.

**Left for you, as they are your calls**

* `revokeAllForUser()` returns `void` while `deleteByUser()` returns the number of rows deleted. That count is useful to a caller, and `delete()` returns one too. Worth returning `int`?
* `ParameterType::STRING` couples the ORM repository to DBAL for something `setParameter()` infers on its own

**On the question nobody answered**

You asked whether to add `revokeAllExpired` before a release went out. v2.1.0 shipped today, so that particular window has closed, but nothing is lost: `RevokeRefreshTokenManagerInterface` is new and unreleased, so 2.2.0 can carry both methods on it at no extra cost in interfaces. If you still want it, it is welcome, here or as a follow up.

And the Psalm failure you flagged is gone. #430 turned out to be a broken step in that workflow, fixed and closed earlier today.

shakaran:
Went with `int` on `revokeAllForUser()`: the repository already reported the number of deleted rows and the manager was dropping it, and `delete()` reports one too. The test asserts the count reaches the caller now.

`ParameterType::STRING` I left as you wrote it. It is explicit rather than wrong, and not worth another round.

Also added the two interfaces to the changelog under Unreleased, since they are new public API that will go out with the next release.

Merging. Thanks for the contribution and for the patience with how long it took to get looked at.

Rainrider:
> Went with int on revokeAllForUser(): the repository already reported the number of deleted rows and the manager was dropping it, and delete() reports one too.

The problem with that is that it complicates the addition of other backends (say Redis) in case they do not return the number of deleted entries. I could also argue about the value of this return to the caller. If I call delete, I want all relevant tokens deleted, I do not care how many they were. I do understand that you value consistency though, so this is fine with me.

> You asked whether to add revokeAllExpired before a release went out. v2.1.0 shipped today, so that particular window has closed, but nothing is lost: RevokeRefreshTokenManagerInterface is new and unreleased, so 2.2.0 can carry both methods on it at no extra cost in interfaces. If you still want it, it is welcome, here or as a follow up.

I will try to add this to the 2.2 branch this week if I find the time.

Thank you for the thorough review and the merge!

PR #430, opened 2026-06-07
ci: fix security scan workflow

Rainrider:
This PR fixes the psalm security workflow.

Not sure what is exactly broken with psalm/psalm-github-actions but it basically runs [this command](https://github.com/psalm/psalm-github-actions/blob/26f175f4d1d9006ea675bb78831ae94126017b07/entrypoint.sh#L89) against the code, so let's just do that directly.

This also uses the version of psalm from composer.json, so the results may be verified locally as well. Caching of composer dependencies could be added as well.

### DIFF
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -32,39 +32,21 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.4'
+          php-version: ${{ matrix.php }}
           extensions: ${{ matrix.extensions }}
 
-      - name: Update Composer dependencies
-        run: composer update
-
-      - name: Install PHP MongoDB extension
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y php8.4-mongodb
-
-      - name: Install Composer dependencies
+      - name: Install Composer Dependencies
         run: composer install
 
-      - name: Mark /github/workspace as safe
-        run: git config --global --add safe.directory /github/workspace
-
       - name: Psalm Security Scan
-        uses: docker://ghcr.io/psalm/psalm-github-actions:6.13.0
-        with:
-          composer_require_dev: true
-          composer_ignore_platform_reqs: true
-          security_analysis: true
-          show_info: true
-          php_version: ${{ matrix.php }}
-          report_file: results.sarif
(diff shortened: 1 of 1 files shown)


shakaran:
Thanks @Rainrider, you were right on both counts, and I have landed your approach in 6dd6f0d.

**What was actually broken**

The job died on the `Install PHP MongoDB extension` step: `php8.4-mongodb` does not exist in the ubuntu-latest repositories, so apt exited with code 100. That step was redundant anyway, since setup-php already installs the extension from the matrix. Removed in e227c7b, which is why the workflow has been green again for a while.

**But your real point still stood**

The workflow installed the psalm declared in `composer.json` and then handed the code to a Docker image that ignored it. Straight from the log of a run that reported success:

```
- Locking vimeo/psalm (7.0.0-beta19)   <- what the composer step installed
...
Psalm 6.13.0@70cdf647...               <- what the container actually ran
       No errors found!
```

So the composer step only existed to have a `vendor/` on disk, and the analysis ran on a different psalm with a config nobody could reproduce locally. Exactly what you described.

**Why I could not merge this as it is**

The repository has no psalm config, and without one Psalm starts at the strictest error level. Running your command against master gives:

```
bin/psalm --force-jit --output-format=github --taint-analysis --show-info=true
→ 220 errors found     EXIT=2
```

All 220 are in `src/`, and mostly level-1 noise: 106 `MissingOverrideAttribute`, 64 `MixedArgument`, 46 `MissingPureAnnotation`, 42 `UnusedClass`. A `UnusedClass` on `ClearInvalidRefreshTokensCommand` gives the flavour: with no config, Psalm cannot tell it is library code. So the job would have gone red.

**What landed**

Your workflow, plus a committed `psalm.xml.dist` to make it pass:

* `bin/psalm` instead of the Docker image, so the results are reproducible locally
* `github/codeql-action/upload-sarif@v4` and `actions/checkout@v6`, as you had them
* the `safe.directory` step dropped, it only existed for the container
* `php-version` from the matrix instead of the hardcoded value
* `errorLevel="8"` and `findUnusedCode="false"` to start from a green baseline, with the annotation checks Psalm 7 added suppressed for now. The plan is to tighten it over time.

I dropped `--force-jit`: it is only a speed flag and opcache is not guaranteed on the runner. I also kept `composer update` rather than `composer install`, since `composer.lock` is gitignored.

The scan now runs the project's own Psalm 7.0.0-beta19 and reports the same 44 informational issues as it does locally, which are a decent to-do list. It also dropped from 1m12s to 30s without the image pull.

Closing this since the commit carries your changes, but the diagnosis was yours. Thanks for digging into it.

PR #432, opened 2026-08-05
feat: add direct expired token revocation

Rainrider:
This PR adds expired token revocation performed directly by the underlying database driver, thus no refresh token objects are added to the unit of work and no life-cycle events are raised. This should be the most performant deletion strategy as no object hydration is performed by doctrine (other than the query result).

## Implementation details:

- `revokeAllExpired` and `deleteAllExpired` are added to their corresponding interfaces using `@method` annotations because the interfaces have been released in 2.2.0 and extending them otherwise would break backwards compatibility. I also didn't want to introduce yet another interface. At some point you might want to deprecate the new interfaces and merge them into the old ones (maybe version 4)


shakaran (changes requested):
Thanks for this — and thanks for picking up the thread from the earlier discussion about `revokeAllExpired`. The reasoning in your description is right on the part that usually gets people into trouble: the interfaces went out in 2.2.0 and adding methods to them would break every implementation. The tests cover all three backends, and CI is green.

There is one thing to change before this can go in, and it is about where rather than what. Then a few smaller ones.

## Please retarget this at `master`

This is the important one. `2.2.x` is the maintenance branch for applications still on Symfony 6.4 and 7.x. **3.0.0 was released today from `master`, and it does not have this method.**

So if this merges into `2.2.x` and goes out, someone upgrading 2.3 → 3.0 *loses* `revokeAllExpired()`. A feature that exists only on the older branch is a regression on the newer one, and it is the kind that surfaces as a fatal error in somebody's console command rather than as anything they were warned about.

There is a second problem underneath that: 2.2.2 is the current release on that branch, and new public API cannot go out in a patch. It would have to be 2.3.0 — a minor from a maintenance branch, which is not what that branch is for.

`master` is where this belongs. If you would also like it available to people who cannot move to Symfony 8, we can port it down to `2.2.x` afterwards as a deliberate backport, and that ordering makes the upgrade path monotonic.

Worth knowing before you rebase: `master` has moved a fair amount since 2.2. Refresh tokens now belong to a family, `revokeFamily()` exists on a new `Model\FamilyRefreshTokenManagerInterface`, and there is a `Cache\CacheRefreshTokenManager` that stores tokens in a PSR-6 pool. That last one matters to you: it implements only what a pool can honour, so whatever shape this lands in has to be something it can decline rather than something it must implement.

## Use a separate interface rather than `@method`

You wrote that you did not want to introduce yet another interface. I understand the instinct, but a separate interface *is* the pattern this bundle uses for exactly this problem — it is why `ListRefreshTokenManagerInterface`, `RevokeRefreshTokenManagerInterface` and `DeleteRefreshTokenRepositoryInterface` are three interfaces instead of one, and 3.0 added two more the same way (`FamilyRefreshTokenManagerInterface`, `DeleteRefreshTokenFamilyRepositoryInterface`).

The practical difference is that an interface can be checked and `@method` cannot. Compare what you have:

```php
if (!method_exists($this->repository, 'deleteAllExpired')) {
    throw new LogicException(...);
}
```

against what the rest of the class does:

```php
if (!$this->repository instanceof DeleteRefreshTokenRepositoryInterface) {
    throw new LogicException(...);
}
```

The second is visible to phpstan and psalm, both of which run over this codebase at their strictest; the first is a runtime string comparison they cannot reason about. It also gives an application a way to ask "can this manager do it?" without calling it and catching an exception, which is what the container needs in order to wire the feature only where it works.

So: `Model\ExpiredRefreshTokenManagerInterface` (or a name you prefer) with `revokeAllExpired()`, and `Doctrine\DeleteExpiredRefreshTokenRepositoryInterface` with `deleteAllExpired()`. Nothing implementing the current interfaces breaks, which was your goal, and the new capability is type-checkable.

## `executeStatement()` rather than `executeQuery()` in the DBAL manager

```php
return (int) $this->connection->createQueryBuilder()
    ->delete($this->quoteTableIdentifier())
    ...
    ->executeQuery()
    ->rowCount();
```

DBAL's own docblocks say it plainly — `QueryBuilder::executeQuery()` is "Executes an SQL query (SELECT)", while `executeStatement()` is "Should be used for INSERT, UPDATE and DELETE" and returns the affected row count directly.

It does work today: I ran it against SQLite and it returns the right number, because most drivers populate `rowCount()` for DML anyway. But it is not the documented contract, it is not portable by DBAL's own warning, and every other write in that file already uses `executeStatement()`. It should be:

```php
return (int) $this->connection->createQueryBuilder()
    ->delete($this->quoteTableIdentifier())
    ->where($this->quoteColumnIdentifier('valid').' < :datetime')
    ->setParameter('datetime', $datetime, Types::DATETIME_MUTABLE)
    ->executeStatement();
```

## The ODM repository is missing its return type

```php
public function deleteAllExpired(DateTimeInterface $datetime = new DateTime())
```

The ORM one has `: int` and the `@method` annotation promises `int`. Worth adding so the two backends have the same signature.

## The names now compete

The bundle already has `revokeAllInvalid()` and `revokeAllInvalidBatch()`, which mean the same thing as `revokeAllExpired()` — a token whose `valid` is in the past. Three methods with two words for one concept is the kind of thing that reads fine while you are writing it and badly a year later.

I think the tell is in your own diff: you had to rename `testRevokeExpiredTokens` to `testRevokeInvalidTokens` to make room for the new test. That is two names fighting over one slot.

I do not have a strong preference for which word wins, but it would be good to settle it. Either align on one ("invalid", since it is what is already published) and let the distinction be carried by the method name — something like `revokeAllInvalidDirectly()` — or keep `expired` and document, at the interface, plainly: this one deletes in the database and returns a count, those two hydrate and return the tokens, and here is when you want each.

## Wire it into `gesdinet:jwt:clear`

This is the part I would most like to see, because it is where the benefit actually lands. Avoiding hydration matters on a table with a large backlog, and clearing a large backlog is precisely what that command is for — but the command still goes through `revokeAllInvalidBatch()`.

One wrinkle to design around rather than ignore: the command reports the revoked tokens with `-v`, and a direct DELETE cannot produce them. So this wants to be a flag rather than a replacement — something like `--no-report`, or letting `-v` keep the hydrating path — and the help text should say what the trade is. Happy to discuss the shape before you write it if you would rather agree it first.

## Small one

StyleCI is failing on trailing whitespace in the docblock lines you added above the `@method` tags.

---

None of this is a criticism of the work, which is solid and well tested. The branch is the only thing I would call blocking; the rest is the kind of tidying that is much easier to do while the change is still fresh. If retargeting at `master` turns out to be more of a rebase than you want to take on, say so and we can work out how to split it.

Rainrider:
## I fixed:

- the style issues (how would I have found those out locally?)
- usage of `executeStatement` instead of `executeQuery`
- missing return type in the ODM repository

## About the rest:

I am sorry but I cannot compete against AI. I waited almost two months for the review of #429 where I stated I would like to add `RevokeRefreshTokenManagerInterface::revokeAllExpired`. And then in 2 days there are 176 commits and 6 releases. And now you ask me to re-target to master, which I won't use, and to backport then to 2.2.x. I cannot justify the time spent on this. Feel free to use, refactor, backport, or just discard this, but it is definitely not worth my time.

Please do not take my words harshly. I do ackwoledge all of those are your calls and it is up to you how you govern your code and how and whether you accept contributions. I just want to share my point of view in case it is of interest.

Please allow me to share my views on the current upgrade path of the bundle.

Symfony has a release cycle of 4 minor versions within a major release with the last one (the .4) being the LTS (long time support). Not supporting the current two LTS releases (6.4 and 7.4) in latest release of this bundle is imho a bad call. Supporting current LTS from the latest release would incentivate me to target rather the latest release than the lowest. Also targeting the lowest release accepting new features is common in both Symfony and Doctrine.

> There is a second problem underneath that: 2.2.2 is the current release on that branch, and new public API cannot go out in a patch. It would have to be 2.3.0 — a minor from a maintenance branch, which is not what that branch is for.

> master is where this belongs. If you would also like it available to people who cannot move to Symfony 8, we can port it down to 2.2.x afterwards as a deliberate backport, and that ordering makes the upgrade path monotonic.

I do not understand this. If this would be backported, than 2.3.0 would be needed too, woundn't it?

I do agree that adding more interfaces is normally the right thing but not at this pace and not in this amount. It becomes confusing at some point and harder to find appropiate names and it delutes the domain language (find vs get vs list, invalid vs. expired). Symony itself extends interfaces with `@method` annotations. The next major then moves from annotations to concrete methods, thus breaking backwards compatibility, but this is what majors are for. Also the point of Psalm and PHPStan not properly supporting `method_exists` is weak imo. If this is considered a failure, it should be reported by CI, but it currenly isn't.

> One wrinkle to design around rather than ignore: the command reports the revoked tokens with -v, and a direct DELETE cannot produce them. So this wants to be a flag rather than a replacement — something like --no-report, or letting -v keep the hydrating path — and the help text should say what the trade is. Happy to discuss the shape before you write it if you would rather agree it first.

This is exacly why I did not update `gesdinet:jwt:clear` to use `revokeAllExpired`. Also `revokeAllExpired` does not raise life-cycle events which current users might depend on. It would have been easier to either add a new console command or deprecate the current one in the 2.x series and remove it in 3.0. Since 3.0 landed today this path has shifted towards 4.0 maybe. Again I lack the time for this much differentiation. My plan was to directly wire the RefreshTokenManager in a user-land service and use that to clean up expired tokens.

Again, in case it comes along like that, I don't mean to dictate or demand anything. You are the maintainer, all of those are your calls and you certainly had good reasons to make them. Thank you a ton for your bundle and for accepting my previous PR, it was my best way to say thank you and try to give something back. I just sadly cannot keep up with the pace from the last 2 days.

